PR #24, opened 2024-10-17
Pin version to be consistent with aiida-core

agoscinski:


.github/workflows/ci.yaml
```diff
@@ -43,10 +43,10 @@ jobs:
     timeout-minutes: 15
     steps:
     - uses: actions/checkout@v4
-    - name: Set up Python ${{ matrix.python-version }}
```
agoscinski, line 46: These were wrong because for docs and formatting we anyway want to use a fixed version

src/wcflow/parsing/_yaml_data_models.py
```diff
@@ -169,9 +169,9 @@ class ConfigCycleTask(_NamedBaseModel):
     To create an instance of a task in a cycle defined in a workflow file.
     """
 
-    inputs: Optional[list[ConfigCycleTaskInput | str]] = Field(default_factory=list)
-    outputs: Optional[list[ConfigCycleTaskOutput | str]] = Field(default_factory=list)
-    depends: Optional[list[ConfigCycleTaskDepend | str]] = Field(default_factory=list)
+    inputs: list[ConfigCycleTaskInput | str] | None = Field(default_factory=list)
```
agoscinski, line 172: changes needed because formatter complained

leclairm (approved):
Thanks!

PR #45, opened 2024-11-26
AiiDA workgraph creation rebased on new IR

agoscinski:
Please merge first  #75

Implementation of running Shell Task with workgraph.

Initialization of all files required to run workflows. The workflow config files are updated to work properly. The $PWD environment is removed because the config_rootdir is now used.

To allow a graph like initialization of workgraph tasks first creating the task nodes and then linking the additional information including inpouts outputs and arguments we implemented a temporary hack to overcome internal validations.

TODO:
- [x] fix tests
- [x] update IR tests
- [x] I think we should upgrade to workgraph 0.4 to avoid headaches later on --> Made issue #76, not high priority
- [x] ~possibility to run the tests with the current profile to debug the workflow? --> No easy solution found for now
- [x] mv files for IR tests into separate folder (in separate PR) --> Not needed since all files are in the config folder

### DIFF
--- a/src/sirocco/workgraph.py
+++ b/src/sirocco/workgraph.py
@@ -1,0 +1,303 @@
+from __future__ import annotations
+
+from pathlib import Path
+from typing import TYPE_CHECKING, Any
+
+import aiida.common
+import aiida.orm
+import aiida_workgraph.engine.utils  # type: ignore[import-untyped]
+from aiida_workgraph import WorkGraph
+
+from sirocco.core._tasks.icon_task import IconTask
+from sirocco.core._tasks.shell_task import ShellTask
+
+if TYPE_CHECKING:
+    from aiida_workgraph.socket import TaskSocket  # type: ignore[import-untyped]
+
+    from sirocco import core
+    from sirocco.core import graph_items
+
+
+# This is hack to aiida-workgraph, merging this into aiida-workgraph properly would require
+# some major refactor see issue https://github.com/aiidateam/aiida-workgraph/issues/168
+# It might be better to give up on the graph like construction and just create the task
+# directly with inputs, arguments and outputs
+def _prepare_for_shell_task(task: dict, kwargs: dict) -> dict:
+    """Prepare the inputs for ShellTask"""
+    from aiida.common import lang
+    from aiida.orm import AbstractCode
+    from aiida_shell.launch import convert_nodes_single_file_data, prepare_code
+
+    command = kwargs.pop("command", None)
+    resolve_command = kwargs.pop("resolve_command", False)
+    metadata = kwargs.pop("metadata", {})
+    # setup code
+    if isinstance(command, str):
+        computer = (metadata or {}).get("options", {}).pop("computer", None)
+        code = prepare_code(command, computer, resolve_command)
+    else:
+        lang.type_check(command, AbstractCode)
+        code = command
+    # update the tasks with links
+    nodes = convert_nodes_single_file_data(kwargs.pop("nodes", {}))
+    # find all keys in kwargs start with "nodes."
+    for key in list(kwargs.keys()):
+        if key.startswith("nodes."):
+            nodes[key[6:]] = kwargs.pop(key)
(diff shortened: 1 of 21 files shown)


GeigerJ2:
OK, @agoscinski, all conflicts should be resolved now. Will branch off from this here, to work on the `ShellTask`.

tests/cases/small/config/test_config_small.yml
```diff
@@ -23,20 +27,24 @@ cycles:
 tasks:
   - icon:
       plugin: shell
-      command: $PWD/tests/files/scripts/icon.py
+      command: scripts/icon.py
       cli_arguments: "{--restart icon_restart} {--init initial_conditions}"
+      env_source_files: data/dummy_source_file.sh
```
agoscinski, line 32: Has been added to test this in the workgraph as we do not run the large config where this is also used

tests/test_wc_workflow.py
```diff
@@ -64,3 +66,18 @@ def test_serialize_workflow(config_paths, pprinter):
...
     VizGraph.from_yaml(config_paths["yml"]).draw(file_path=config_paths["svg"])
+
+
+# configs that are tested for running workgraph
+@pytest.mark.parametrize(
+    "config_path",
+    [
+        "tests/cases/small/config/test_config_small.yml",
+        "tests/cases/parameters/config/test_config_parameters.yml",
+    ],
```
agoscinski, line 77: the large config takes 4 minutes to run therefore it is not included in the tests

tests/conftest.py
```diff
@@ -0,0 +1 @@
+pytest_plugins = ["aiida.tools.pytest_fixtures"]
```
agoscinski, line 1: This autoloads a temporary aiida profile for the tests. I find pytest is doing here a bit too much in the background. One needs to look into the auto fixtures from aiida to see what is all created. Just keep in mind that the tests are not run with your current profile but create a temporary one.

tests/test_wc_workflow.py
```diff
@@ -64,3 +66,18 @@ def test_serialize_workflow(config_paths, pprinter):
...
+)
+def test_run_workgraph(config_path):
```
agoscinski, line 79: TODO write a comment that this uses automatically created aiida profile, so to debug this you want to run it outside the tests

src/sirocco/workgraph.py
```diff
@@ -0,0 +1,313 @@
...
+            for task in cycle.tasks:
+                self._link_wait_on_to_task(task)
+
+        for cycle in self._core_workflow.cycles:
+            for task in cycle.tasks:
+                for input_ in task.inputs:
+                    self._link_input_nodes_to_task(task, input_)
+                self._link_arguments_to_task(task)
+                for output in task.outputs:
+                    self._link_output_nodes_to_task(task, output)
```
agoscinski, line 179: directly iterate through `self._core_workflow.tasks`

agoscinski: Move the for loop for the outputs before the inputs.

src/sirocco/workgraph.py
```diff
@@ -0,0 +1,313 @@
...
+        "outputs": list(missing_outputs),
+        "parser": kwargs.pop("parser", None),
+        "metadata": metadata or {},
+    }
+
+
+aiida_workgraph.engine.utils.prepare_for_shell_task = _prepare_for_shell_task
+
+
+class AiidaWorkGraph:
```
agoscinski, line 67: `WorkgraphConverter` a staticmethod 

Maybe inherit from the IR workflow and just add a function `AiiDAWorkflow`

agoscinski: I will do this in a separate PR as this touches so much of the code that rebasing will become painful

agoscinski: See issue https://github.com/C2SM/Sirocco/issues/82

leclairm (approved):
just approved so that we can continue but I'm not into AiiDA WorkGraph and AiiDA internals so I can't really judge.

PR #56, opened 2024-11-29
Fuse classes for wait_on tasks and input data

leclairm:
Fuse classes for wait_on tasks and input data as they use the exact same keywords for the same purpose, namely targeting other nodes when building the graph.

Also reorder the classes in `_yaml_data_models.py` so that they correspond more to the order in the example `yaml` files (even though this order is just a convention and can be changed by the users without consequences)

### DIFF
--- a/src/sirocco/parsing/_yaml_data_models.py
+++ b/src/sirocco/parsing/_yaml_data_models.py
@@ -71,7 +71,7 @@ class _WhenBaseModel(BaseModel):
 
 
 class TargetNodesBaseModel(_NamedBaseModel):
-    """class for targetting other task or data nodes in the graph"""
+    """class for targeting other task or data nodes in the graph"""
 
     model_config = ConfigDict(arbitrary_types_allowed=True)
     date: list[datetime] = []  # this is safe in pydantic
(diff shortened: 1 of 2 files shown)


src/sirocco/parsing/_yaml_data_models.py
```diff
@@ -70,10 +70,8 @@ def convert_datetime(cls, value) -> datetime:
         return datetime.fromisoformat(value)
 
 
-# TODO: Change class name, does not fit anymore wit hthe addition of `when` and `parameters`
-#       find something more related to graph specification in general like _GraphTargetBaseModel
-class _LagDateBaseModel(BaseModel):
-    """Base class for all classes containg a list of dates or time lags."""
+class TargetNodesBaseModel(_NamedBaseModel):
+    """class for targetting other task or data nodes in the graph"""
```
agoscinski, line 74: Can you explain this a bit more in the docstring? Reading it, it is not so clear for me

agoscinski: Maybe something like this. But I am also not really happy with this
```suggestion
    """Class for specifying time information relative to the cycle date, and
    information related to the parametrization. """
```
I hope when we move to a release things become clearer

agoscinski (approved):

PR #58, opened 2024-12-02
Fixes pygraphviz failure when building wheels

agoscinski:
One needs to install graphviz over the system package manager before installing pygraphviz to prevent an automatic build that fails. This can is achieved widh `sudo apt-get install graphviz graphviz-dev`.

We also add the `workflow_dispatch` event so we can run the workflow manually.

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
     - main
+  workflow_dispatch:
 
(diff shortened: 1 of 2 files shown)


agoscinski:
This run worked https://github.com/C2SM/Sirocco/actions/runs/12120717931/job/33790028069?pr=58

leclairm (approved):
I blindly trust you on this! :wink:

PR #63, opened 2024-12-12
Split parsing classes between declaration and validation

leclairm:
## split Task and Data classes in `parsing` in 2 classes:
- one for declaration of fields that are needed in core classes
- one for pydantic validation

This way core classes can inherit from the declaration classes and we avoid duplication.

## remove metaclass for `core.graph_items.Task`

this was overkill, replace by `__init_subclass__`

### DIFF
--- a/src/sirocco/parsing/_yaml_data_models.py
+++ b/src/sirocco/parsing/_yaml_data_models.py
@@ -3,15 +3,15 @@ from __future__ import annotations
 import time
 from datetime import datetime
 from pathlib import Path
-from typing import Annotated, Any, Literal
+from typing import Annotated, Any, ClassVar, Literal
 
 from isoduration import parse_duration
 from isoduration.types import Duration  # pydantic needs type # noqa: TCH002
 from pydantic import BaseModel, ConfigDict, Discriminator, Field, Tag, field_validator, model_validator
 
-from sirocco.core._tasks.icon_task import IconTask
-from sirocco.core._tasks.shell_task import ShellTask
-from sirocco.core.graph_items import Task
+# from sirocco.core._tasks.icon_task import IconTask
+# from sirocco.core._tasks.shell_task import ShellTask
+# from sirocco.core.graph_items import Task
 from sirocco.parsing._utils import TimeUtils
 
 
@@ -251,7 +251,8 @@ class ConfigBaseTask(_NamedBaseModel):
 
     # this class could be used for constructing a root task we therefore need a
     # default value for the plugin as it is not required
-    plugin: Literal[Task.plugin] | None = None
+    # plugin: Literal[Task.plugin] | None = None
+    plugin: ClassVar[Literal["_BASE_TASK_"]] = "_BASE_TASK_"
     parameters: list[str] = Field(default_factory=list)
     host: str | None = None
     account: str | None = None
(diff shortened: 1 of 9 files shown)


src/sirocco/parsing/_yaml_data_models.py
```diff
@@ -244,21 +242,27 @@ def check_period_is_not_negative_or_zero(self) -> ConfigCycle:
...
     """
 
     # this class could be used for constructing a root task we therefore need a
     # default value for the plugin as it is not required
-    # plugin: Literal[Task.plugin] | None = None
-    plugin: ClassVar[Literal["_BASE_TASK_"]] = "_BASE_TASK_"
+    # WORKAROUND for the mixin of _NamedBaseModel subclasses and dataclasses:
+    # use base_plugin instead of plugin in order to not overrite
+    # the plugin value provided by in ConfigXxxTaskCore
+    base_plugin: ClassVar[Literal["_BASE_TASK_"]] = "_BASE_TASK_"
```
agoscinski, line 264: Can't you just shuffle? the inheritance so the plugin of the customized plugin is prioritized?
```
ConfigShellTaskCore(ConfigBaseTask, ConfigShellTaskCore) 
-> ConfigShellTask(ConfigShellTaskCore, ConfigBaseTask)
```
see
```
class A:
    plugin = 5


class B:
    plugin = 6


class C(A,B):
    pass


print(C.plugin) # out 5

class C(B,A):
    pass

print(C.plugin) # out 6
```

agoscinski (dismissed):
Just would like to avoid to use `Core` due to our core module. Would be one of these okay `ConfigBaseTaskSpecs`, `ConfigIconBaseCommon`, `ConfigBaseTaskStruct`, `ConfigBaseTaskDataClass`?

leclairm:
> Just would like to avoid to use `Core` due to our core module. Would be one of these okay `ConfigBaseTaskSpecs`, `ConfigIconBaseCommon`, `ConfigBaseTaskStruct`, `ConfigBaseTaskDataClass`?

Yes, I didn't have a strong opinion there. I wanted to express that the class contains the part that will be reused in core, which is everything except parameters actually. I'd be fine with `ConfigBaseXxxSpecs`.

src/sirocco/parsing/_yaml_data_models.py
leclairm: I cannot change the order here. I tried but it fails because of the MRO. The dataclass first needs the `_NamedBaseModel.merge_name_and_specs` to do its job before assigning its own attributes. That's why I renamed `ConfigBaseTask.plugin` as `ConfigBaseTask.base_plugin` to avoid the overwrite.

agoscinski: But can't you change the order by overwriting the the __init__ constructor?
```python
_NamedBaseModel.__init__(self, ...)
ConfigBaseTask.__init__(self, ...)
```

GeigerJ2 (approved):
LGTM. Nothing more to add from my side!

src/sirocco/parsing/_yaml_data_models.py
agoscinski: Maybe the plugin just does not make sense to have in the base class, we don't really want to ever create a base class except in the root class case. So maybe it makes more sense to create a dedicated root class

https://github.com/C2SM/Sirocco/commit/b47ee844847990d3851d74dfc95e2d0ab5836bfc
I deleted also the `__init__`constructor from the root class because it does not make sense with the current code anymore. We need to implement this properly at some other point.

leclairm: Thanks! It makes more sense like this, I merged it. And yes, we'll have to implement this root thing at some point but no hurry. I?ll make it an issue for the first release milestone.

PR #66, opened 2024-12-16
Arbitrary CLI args for shell tasks

GeigerJ2:
This PR adds the `_CliArgsBaseModel` class (inspired from `_WhenBaseModel`), which replaces the `command_option` and `input_arg_options` of the `ConfigShellTaskSpecs`. Validation is applied on the correctness of the keyword and positional arguments to ensure they start with `-` or `--`.

I updated the three test YAML files accordingly. This change led to changes in the pretty-print test text files, which is also part of the PR. @ all: Currently, they all appear on one line. Not sure if formatting should be done nicer here. I saw that the `self.as_block` is only applied for entities of the `ConfigCycleTask`, so I just left the `.txt` files as I got them from running the test and updated them in the PR.

Actually making something useful out of these arguments only happens when the WG is created, so will require #45 to be merged.


src/sirocco/parsing/_yaml_data_models.py
```diff
@@ -82,6 +82,29 @@ def convert_datetime(cls, value) -> datetime:
...
 
 
+class _CliArgsBaseModel(BaseModel):
+    """Base class for cli_arguments specifications"""
+
+    # TODO: Even allow for `str`, or always require list?
+    positional: str | list[str] | None = None
+    # Field needed for child class doing pydantic parsing
+    keyword: dict[str, str] | None = Field(default_factory=dict)
+    flags: str | list[str] | None = None
```
leclairm, line 92: I'm wondering if these should really be 2 different categories. Couldn't flags be keywords without a corresponding argument?

src/sirocco/parsing/_yaml_data_models.py
```diff
@@ -82,6 +82,29 @@ def convert_datetime(cls, value) -> datetime:
...
+    source_file: str | list[str] | None = None
+
+    # TODO: Should we validate here, or allow users to pass it without the hyphen(s), and prepend them automatically?
+    @field_validator("keyword", mode="before")
+    @classmethod
+    def validate_keyword_args(cls, value):
```
leclairm, line 98: This should also validate flags if they are not merged with keywords

tests/files/configs/test_config_large.yml
```diff
@@ -74,10 +75,14 @@ tasks:
...
       command: $PWD/examples/files/scripts/cleanup.sh
-      input_arg_options:
-        grid_file: '-g'
-        extpar_file : '-p'
-        ERA5: '-e'
+      cli_arguments:
+        positional:
+          - grid_file
+        keyword:
+          -g: grid_file
```
leclairm, line 82: I know we don't care yet about how much sense those arguments have but we could still remove one of the two `grid_file` arguments

leclairm:
> I updated the three test YAML files accordingly. This change led to changes in the pretty-print test text files, which is also part of the PR. @ALL: Currently, they all appear on one line. Not sure if formatting should be done nicer here. I saw that the `self.as_block` is only applied for entities of the `ConfigCycleTask`, so I just left the `.txt` files as I got them from running the test and updated them in the PR.

At some point, we might have to differentiate between pretty printing for users and for serialized test data. For now, I'd say we keep it his way as testing is more important. Later we could work on stg like a verbosity level and other pretty printing options for users as opposed to testing.

src/sirocco/parsing/_yaml_data_models.py
GeigerJ2: Hm, interesting idea. I don't have a strong opinion, though, I do like to have things clean and separate, usually. Maybe @agoscinski can also weigh in?

src/sirocco/parsing/_yaml_data_models.py
```diff
@@ -82,6 +82,29 @@ def convert_datetime(cls, value) -> datetime:
...
+    """Base class for cli_arguments specifications"""
+
+    # TODO: Even allow for `str`, or always require list?
+    positional: str | list[str] | None = None
+    # Field needed for child class doing pydantic parsing
+    keyword: dict[str, str] | None = Field(default_factory=dict)
+    flags: str | list[str] | None = None
+    source_file: str | list[str] | None = None
+
+    # TODO: Should we validate here, or allow users to pass it without the hyphen(s), and prepend them automatically?
```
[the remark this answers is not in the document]
GeigerJ2: Yeah, I think I forgot to remove that, as in the end, I'm validating. And the idea of letting the user pass it without `-` or `--` might be a bad one ^^

src/sirocco/parsing/_yaml_data_models.py
GeigerJ2: Fair enough! Will add it!

tests/files/configs/test_config_large.yml
GeigerJ2: Will do!

src/sirocco/parsing/_yaml_data_models.py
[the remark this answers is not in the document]
GeigerJ2: Done!

tests/files/configs/test_config_large.yml
GeigerJ2: Done!

GeigerJ2:
Thanks for the quick review, @leclairm! I addressed your points. For the idea of specifying flags as keywords without values, maybe we can keep that in mind for the future, and advance as is for now? I'd like to continue with the WorkGraph creation this afternoon, so that these CLI arguments actually get resolved to something meaningful. So it would be nice to get this merged before.

> At some point, we might have to differentiate between pretty printing for users and for serialized test data. For now, I'd say we keep it his way as testing is more important. Later we could work on stg like a verbosity level and other pretty printing options for users as opposed to testing.

Fully agree!

leclairm (approved):
all good, keep merging kw args and flags for later

PR #67, opened 2024-12-16
Fix svgs being created in main directory.

GeigerJ2:
Will update base branch when #66 is merged.


tests/test_wc_workflow.py
leclairm, line 1: No need for intermediate variables like `config_path` and `reference_path` with the new dict.

.gitignore
```diff
@@ -164,3 +164,4 @@ cython_debug/
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
 #.idea/
 
+tests/files/svgs/*.svg
```
leclairm, line 167:
```suggestion
tests/files/svgs/*
```

GeigerJ2: Wasn't sure if this then also ignores the `.gitkeep` file I had to put in the directory, so that it gets tracked by git (which empty dirs usually don't ^^)

tests/test_wc_workflow.py
GeigerJ2: Ah, actually true :D wanted to try and keep the previous implementation as unchanged as possible.

.gitignore
GeigerJ2: OK, I just tested this and if I exclude _all_ files like in your suggestion, then `.gitkeep` has no effect, which means the `svgs` directory is not under version control and doesn't get synced. However, with the directory missing, the tests fail, as they expect this directory to write the `svgs`. So I'd leave it as is for now.

leclairm: Ah true, sorry, I typed a bit too fast.
```
tests/files/svgs/*
!tests/files/svgs/.gitkeep
```

tests/test_wc_workflow.py
leclairm: Only `reference_path` is reused once. I think we can afford the overhead of reading it from the dict twice (which is probably 0 as it is probably still in the cache) and get rid of the additional line.

.gitignore
GeigerJ2: I see, makes sense!

tests/test_wc_workflow.py
GeigerJ2: In terms of the dictionary access, it's not about the overhead for me, but rather about re-use. If we'd change the key to, say `".txt"` instead of `"txt"` (stupid example, but just for illustration), we would otherwise have to update that in two places. Whereas, if we assign it to a variable, only in one place. Though, I don't think it's important right now, so I'll update it.

leclairm: That would be an argument to not use a dict at all :stuck_out_tongue_closed_eyes: and just add an element to the tuple, expanding it with
```python
config_path, reference_path, _ = config_case 
```
if needed

GeigerJ2:
Thanks for the quick review, @leclairm! I implemented your proposed changes. Not sure why this PR still shows as 28 commits, despite merging `origin/main` into it, and changing the base branch of the PR, though, that shouldn't matter anyway once I squash-merge.

leclairm (approved):

tests/test_wc_workflow.py
GeigerJ2: I still prefer dict because it's more readable :p

PR #69, opened 2024-12-18
WIP: Shell tasks for WorkGraph with pinned old version

GeigerJ2:
For easy diff


agoscinski:
Should be now completely merged over to #45

PR #71, opened 2024-12-19
Make tasks aware of config context

leclairm:
make tasks aware of config root directory, start and end dates.
Also restructure tests to put each in its own directory


agoscinski (approved):
I already approve. Only one question I am not sure why the tasks have start and end_date? Isn't this something icon specific and should be in the icon task?

PR #74, opened 2025-01-01
Reduce number of iterations of parameters config

agoscinski:
We do this as preparation to decrease the time it takes to run the workflow for this config.


leclairm:
By doing this we totally ignore the yearly cycle. I mean it's fine if we need this for technical reasons for certain tests but in that case I'd remove the yearly cycle as well or create a new test without it. We don't need to activate all the configs for all the tests.

leclairm (approved):

agoscinski:
Because the testing time is still not ideal but acceptable I made an issue to further improve it https://github.com/C2SM/Sirocco/issues/81

PR #75, opened 2025-01-01
Replace `cli_arguments` in the yaml to a single string

agoscinski:
First merge PR #74

When implementing the suggestion from CLI arguments I realized that there is a problem with the association of an option to an argument for input arguments. From the string
`--init {file}` we do not know if the option `--init` is associated with the `file` input or not. This is important since we have time conditions specified with `when` so we can omit certain inputs during a certain time. In the case the option is associated with the input we want to omit both. My solution was to include the option to the string interpolation `{--init file}`. However this makes parsing more cumbersome since we have now a space insinde the curly brackets. We use spaces to split the cli argument into separate entities. In this case we however want both to be together. We also have to consider CLI arguments that use curly brackets for environmental variables `${VAR}_EXTENDED`. For now I implemented In PR #45 that all data inputs have to be specified in the `cli_arguments`. The remaining data inputs which are not specified in the `cli_arguments` are always added as positional arguments in the same order as specified in the yaml file.

### DIFF
--- a/src/sirocco/parsing/_yaml_data_models.py
+++ b/src/sirocco/parsing/_yaml_data_models.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import time
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from datetime import datetime
 from pathlib import Path
 from typing import Annotated, Any, ClassVar, Literal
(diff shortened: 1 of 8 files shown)


src/sirocco/parsing/_yaml_data_models.py
```diff
@@ -306,16 +269,109 @@ class ConfigRootTask(ConfigBaseTask):
...
+    ShellCliArguments ShellCliArgument(name="file", references_data_item=False, cli_option_of_data_item=None),
+    ShellCliArgument(name="--init", references_data_item=False, cli_option_of_data_item=None),
+    ShellCliArgument(name="file", references_data_item=True, cli_option_of_data_item=None),
+    ShellCliArgument(name="file", references_data_item=True, cli_option_of_data_item="--init")
+
+    Attributes:
+        name: Name of the argument. For the examples it is "file", "--init", "file" and "file"
+        references_data_item: Specifies if the argument references a data item signified by enclosing it by curly
+            brackets.
+        cli_option_of_data_item: The CLI option associated to the data item.
```
agoscinski, line 287: Its not supper nice name but `option` was a but confusing because keep in mind that in case the argument is `--proc=4` then this one argument with name `--proc=4` as we only need the separation of parameter and option for when data items are specified.

leclairm:
Maybe I'm missing something but I don't see the need for splitting into arguments anymore. can't we keep the command string as is and stupidly find/replace what is between curly brackets once we know the path? Because it conflicts with the meaning of `{xxx}` in shell interpreters, we could have placeholders delimited by doubled curly brackets like this `"-D --CMAKE_CXX_COMPILER=${CXX_COMPILER} --init {{file}}"`.

Also the current implementation would fail with `"--init {file}"`, which I assume a lot of users would do.

leclairm:
Which choice do we make when several lags or dates or specified, like in the `large` config?
```yaml
cycle:
[...]
  - yearly:
      start_date: *root_start_date
      end_date: *root_end_date
      period: 'P1Y'
      tasks:
        - postproc_2:
            inputs:
              - stream_2:
                  lag: ['P0M', 'P2M', 'P4M', 'P6M', 'P8M', 'P10M']
            outputs: [postout_2]
[...]
tasks:
  - postproc_2:
      plugin: shell
      command: $PWD/examples/files/scripts/main_script_atm.sh
      cli_arguments: "{--input stream_2}"
      [...]
[...]
```
We can parse it either as
```
--input stream_2_P0M stream_2_P2M stream_2_P4M stream_2_P6M stream_2_P8M stream_2_P10M
```
or
```
--input stream_2_P0M --input stream_2_P2M --input stream_2_P4M --input stream_2_P6M --input stream_2_P8M --input stream_2_P10M
```
where `stream_2_PXXM` is the actual data path when it's known at runtime.

agoscinski:
I have never seen the second case. But can we do this in a future PR? I can make an issue for it and work on it later but it seems to be not essential for the demo.

agoscinski:
We decided that the user has to do this use case explicitly 
```shell
--input stream_2_P0M --input stream_2_P2M --input stream_2_P4M --input stream_2_P6M --input stream_2_P8M --input stream_2_P10M
```
by creating a data item for each period. TODO create issue

leclairm (approved):
Ok for me until we use ports

PR #77, opened 2025-01-03
Upgrade to aiida-workgraph v0.4.10

agoscinski:
Needs to merged after PR #45

To include several bugfixes fwe upgrade to the version v0.4.10. This is required to specify computer in the metadata of the workgraph task as this is not possible with the currently pinned version. The workgraph API has changed therefore we adapt to new API the functions which makes part of the function calls cleaner. The workaround is still needed but has been better documented. Closes issue #76

PR #78, opened 2025-01-03
Add option to specify the computer for a task

agoscinski:
Requires #77 to be merged before.

The user can specify a computer for a task by the aiida label. A computer be initialized through the verdi CLI or programatically. This is for now the only feasible solution to support the whole scope of connections ssh and firecrest without implementing our own CLI resolution logic.


leclairm:
I'm fine with it but not an AiiDA expert so I let @GeigerJ2 and @DropD review it.

agoscinski:
@leclairm approved on slack

PR #79, opened 2025-01-03
Add option to specify computer for available data items

agoscinski:
Needs #78 to be merged before.

This allows to specify data items that are only remotely available or should not be copied over to the aiida internal folder. I did not add the computer information to the IR because the IR focuses on representing the unrolling/expansion from the config to a graph and the computer information is completely independent from this logic.


leclairm:
Same as for #78 , I'm fine with it but not an AiiDA expert so I let @GeigerJ2 and @DropD review it.

agoscinski:
@leclairm approved on slack

PR #83, opened 2025-01-09
Testing string interpolation for multi cli args

agoscinski:
So far the tests only test the support of one argument in the cli args. By using `analysis_foo_bar` which will be splitted into multiple arguments due to the lags, we test now such this use case.

### DIFF
--- a/src/sirocco/parsing/_yaml_data_models.py
+++ b/src/sirocco/parsing/_yaml_data_models.py
@@ -390,6 +390,7 @@ class ConfigBaseDataSpecs:
     type: str | None = None
     src: str | None = None
     format: str | None = None
+    computer: str | None = None
 
 
 class ConfigBaseData(_NamedBaseModel, ConfigBaseDataSpecs):
@@ -417,7 +418,13 @@ class ConfigAvailableData(ConfigBaseData):
 
 
 class ConfigGeneratedData(ConfigBaseData):
-    pass
+    @field_validator("computer")
+    @classmethod
+    def invalid_field(cls, value: str | None) -> str | None:
+        if value is not None:
+            msg = "The field 'computer' can only be specified for available data."
+            raise ValueError(msg)
+        return value
 
 
 class ConfigData(BaseModel):
--- a/src/sirocco/parsing/_yaml_data_models.py
+++ b/src/sirocco/parsing/_yaml_data_models.py
@@ -244,6 +244,7 @@ class ConfigCycle(_NamedBaseModel):
 
 @dataclass
 class ConfigBaseTaskSpecs:
+    computer: str | None = None
     host: str | None = None
     account: str | None = None
     uenv: dict | None = None
--- a/src/sirocco/workgraph.py
+++ b/src/sirocco/workgraph.py
@@ -18,47 +18,51 @@ if TYPE_CHECKING:
     from sirocco.core import graph_items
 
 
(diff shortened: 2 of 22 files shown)


agoscinski:
I see the test pass because it is correctly linked in workgraph, but not in the arguments. To make the tests fail an argument check needs to be done. If one checks for the process `merge.py` the `_aiidasubmit.sh` then one sees the difference before
```shell
'/Users/alexgo/code/Sirocco/tests/cases/parameters/config/scripts/merge.py' 'analysis'  > 'stdout' 2> 'stderr'
```
and after the fix
```shell
'/Users/alexgo/code/Sirocco/tests/cases/parameters/config/scripts/merge.py' 'analysis' 'analysis'  > 'stdout' 2> 'stderr'
```
In the second case one can see that the two inputs for the lags have been added to the arguments. This reveals actually a problem with aiida-shell that I was not aware, that the names are not named uniquely. Have to think about this.

leclairm:
I think GH failed when this happened: Base automatically changed from `remote-data` to `main`. For instance the whole `workgraph.py` is shown as new in the diff.

leclairm:
> I think GH failed when this happened: Base automatically changed from `remote-data` to `main`. For instance the whole `workgraph.py` is shown as new in the diff.

Ah sorry, it's probably because there are still merge conflicts. I guess from rebasing on `main` while you already had all the workgraph part in your branch.

src/sirocco/parsing/_yaml_data_models.py
```diff
@@ -416,7 +418,13 @@ class ConfigAvailableData(ConfigBaseData):
 
 
 class ConfigGeneratedData(ConfigBaseData):
-    pass
+    @field_validator("computer")
+    @classmethod
+    def invalid_field(cls, value: str | None) -> str | None:
```
DropD, line 423:
```suggestion
    def invalid_field(cls, value: str | None) -> None:
```

This never returns anything but `None`.

src/sirocco/workgraph.py
```diff
@@ -0,0 +1,332 @@
...
+        # stores the input data available on initialization
+        self._aiida_data_nodes: dict[str, aiida_workgraph.orm.Data] = {}
+        # stores the outputs sockets of tasks
+        self._aiida_socket_nodes: dict[str, TaskSocket] = {}
+        self._aiida_task_nodes: dict[str, aiida_workgraph.Task] = {}
+
+        self._add_available_data()
+        self._add_tasks()
+
+    def _validate_workflow(self):
```
DropD, line 91: missing type hints

src/sirocco/workgraph.py
```diff
@@ -0,0 +1,332 @@
...
+            self._aiida_data_nodes[label] = aiida.orm.RemoteData(remote_path=input_.src, label=label, computer=computer)
+        elif input_.type == "file":
+            self._aiida_data_nodes[label] = aiida.orm.SinglefileData(label=label, file=input_full_path)
+        elif input_.type == "dir":
+            self._aiida_data_nodes[label] = aiida.orm.FolderData(label=label, tree=input_full_path)
+        else:
+            msg = f"Data type {input_.type!r} not supported. Please use 'file' or 'dir'."
+            raise ValueError(msg)
+
+    def _add_tasks(self):
```
DropD, line 164: missing return type hint

agoscinski:
Has been mostly resolved in #125. There is still an issue #140 but it is not that urgent for early user milestone

Issue #88, opened 2025-01-14
More Mature Datamodels


DropD:
## Problem

The typing in pydantic models created from user config obviously allows for configs to exist which can not successfully be turned into a workflow. In addition, `None` is often used as a default value, when there seems to be a deeper meaning to "information not provided in the yaml file". Some of the downsides of this are:

- Looking at the models it is not immediately clear which information is required, has to come from the user, can be read from elsewhere
- Looking at the models it is not immediately clear what it means if certain values are (apparently) legitimately `None`. For example, does `period=None, start_date != end_date != None` mean "run at start date" or "never run" or "run at start and end date"?
- Any code that operates on instances of models has to check for a whole load of `None` values if not more possible types for static analysis to be able to confirm it's not doing anything wrong, while many of those checks are not necessary, because validators have already taken care of it.
- It is very hard to infer from the code what a meaningful unit test should be.

All of these things block us from making use of unit tests and static type analysis, which are major factors to keep up velocity in the mid-to-long term.

## Proposed solution:

I propose to solve the problem incrementally by doing the following for each of the data models:

1) write a doctest example which demonstrates how to obtain the class from yaml
2) write more extensive unit tests for the desired state after all validators have run
3) add unit tests for how they are used in `workflow.Workflow` creation
4) ask the hard questions about whether `None` or an empty container is meaningful in attributes and if yes, what the meaning is.
5) insert more meaningful sentinel values (enums, types) and additional validators as required
6) insert additional "finalized" data models and canonicalization functionality if validators are not enough
7) test everything that was added
8) switch `workflow.Workflow` building to the new canonicalized types and adapt the unit tests

If at any point any of this requires changes to the existing integration tests, everyone should agree that it is necessary.

agoscinski:
> switch workflow.Workflow building to the new canonicalized types and adapt the unit tests

Was this regarding making AiidaWorkflow to a Builder utils with only staticmethods?

PR #89, opened 2025-01-14
Tests and canonicalization for ConfigWorkflow

DropD:
First step towards clarifying and testing the data layout from which we can reliably build workflows.

## Problem

See #88. This covers steps 1-8 for `ConfigWorkflow`

## Changes

- Doctests
  - enabled by default (required fixing some pretty_printer doctests)
  - Doctests for ConfigWorkflow, outlining the smallest possible valid yaml snippet
- Unit tests
  - Check that we can build a CanonicalWorkflow from code as well as from the mimimal ConfigWorkflow instance
  - Check that we can use the intended `load_workflow_config` method to load the minimal yaml snippet
  - Check that we can convert the minimum valid CanonicalWorkflow into a core.Workflow


tests/unit_tests/core/test_workflow.py
```diff
@@ -0,0 +1,23 @@
...
+            available=[models.ConfigAvailableData(foo={})],
+            generated=[models.ConfigGeneratedData(bar={})],
+        ),
+    )
+
+    testee = workflow.Workflow(minimal_config)
+
+    pretty_print.PrettyPrinter().format(testee)
+
+    assert testee.name is None
```
DropD, line 20: Question: What does it mean to have a `core.Workflow` with `name=None`, `rootdir=None`?

tests/unit_tests/parsing/test_yaml_data_models.py
```diff
@@ -0,0 +1,41 @@
+import textwrap
+
+from sirocco.parsing import _yaml_data_models as models
+
+
+def test_workflow_test_internal_dicts():
+    testee = models.ConfigWorkflow(
+        cycles=[],
+        tasks=[{"some_task": {"plugin": "shell"}}],
```
DropD, line 9: Question: is it a problem, that we can not currently (for testing purposes) pass actual `ConfigXyzTask` instances into the constructor of `ConfigWorkflow`, but have to go through this dictionary form?

tests/unit_tests/core/test_workflow.py
DropD: Realistically this can not happen in normal usage, (as tested in `test_yaml_data_models.py`). However, downstream code has to deal with the possibility or negate the help of static type checking.

tests/unit_tests/parsing/test_yaml_data_models.py
agoscinski: I think yes. As you pointed out in the last meeting, we need to decouple the components for effective testing so we need to test `YAML <-> CONFIG` and `CONFIG <-> GRAPH` separately which requires a creation of the config objects without a yaml string.

DropD: What do you think of the drafted solution in #90, splitting into a "canonical" and a "yaml-parsed" class? Note that 
a) Some of this would be alleviated by idempotent validators, which that approach would introduce gradually
b) The full benefits would only kick in, once all the config classes have undergone the same process. However, testability will increase incrementally.

agoscinski: made a comment there about the canonical workflow

src/sirocco/parsing/_yaml_data_models.py
```diff
@@ -387,6 +395,39 @@ def get_plugin_from_named_base_model(data: dict) -> str:
...
+            ...     '''
+            ... )
+            >>> wf = pydantic_yaml.parse_yaml_raw_as(ConfigWorkflow, config)
+
+        minimum programmatically created instance
+
+            >>> empty_wf = ConfigWorkflow(cycles=[], tasks=[], data={})
+
+    """
+
```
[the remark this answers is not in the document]
agoscinski: we agreed in the meeting to produce an error, maybe it is easier for merging to do this in a subsequent PR

tests/unit_tests/core/test_workflow.py
agoscinski: So we need to check if they are not None in the core.Workflow right? Should we make an issue to keep in mind?

agoscinski (approved):
The questions you point out would need some further changes to be implement but maybe we implement them in another PR to avoid merge conflicts and focus on the canonicalization? I approve already.

leclairm (approved):

tests/unit_tests/core/test_workflow.py
DropD: Ok, I will adapt this in #90

src/sirocco/parsing/_yaml_data_models.py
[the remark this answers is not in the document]
DropD: once I merge #90 into this, the error would be raised during canonicalization

tests/unit_tests/core/test_workflow.py
DropD: I will need to think about how to adapt the canonicalize function to allow passing in additional information though, because the additional information will be different per class if any (not sure if singledispatch is still the right match then).

src/sirocco/parsing/_yaml_data_models.py
```diff
@@ -536,7 +571,48 @@ def check_parameters(self) -> ConfigWorkflow:
...
+    @property
+    def data_dict(self) -> dict[str, ConfigAvailableData | ConfigGeneratedData]:
+        return {data.name: data for data in itertools.chain(self.data.available, self.data.generated)}
+
+    @property
+    def task_dict(self) -> dict[str, ConfigTask]:
+        return {task.name: task for task in self.tasks}
+
+
+def canonicalize_workflow(value: ConfigWorkflow, rootdir: Path) -> CanonicalWorkflow:
```
leclairm, line 601: I guess `value` was the variable name for the `@singledispatch` version. Maybe a more explicit name would be better, like `config_workflow` or a shorter version of it.

PR #92, opened 2025-01-16
Add computer to `Data` initialization in core

agoscinski:
The `computer` argument from the config class was not correctly passed to the core `Data` class. This is fixed here.

Note that aiida-shell has a bug when using `RemoteData`. This will be fixed here https://github.com/sphuber/aiida-shell/pull/107

### DIFF
--- a/src/sirocco/core/graph_items.py
+++ b/src/sirocco/core/graph_items.py
@@ -40,6 +40,7 @@ class Data(ConfigBaseDataSpecs, GraphItem):
     def from_config(cls, config: ConfigBaseData, coordinates: dict) -> Self:
         return cls(
             name=config.name,
+            computer=config.computer,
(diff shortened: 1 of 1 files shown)


agoscinski:
This PR probably fails because localhost is not created as computer. @leclairm will add his contribution to pass the `Config*` object as dictionary in core so we can automatically pass the shared `Config*Spec` arguments. This makes such undetected passed parameters less likely to happen.

TODO: need to use aiida localhost fixture for this

GeigerJ2:
Note to self, try out if https://github.com/sphuber/aiida-shell/pull/108 fixes this.

GeigerJ2:
Moved this change to #136 adding you as co-author. Hope that's OK, @agoscinski!

PR #95, opened 2025-01-20
Remote execution of scripts

agoscinski:
At the moment a code is automatically created from the script file. This works for local execution since the script is available on the computer. For remote execution the script is not available and result into an error.

To copy the script over to the remote machine one could specify the file as remote data with symlink to false. This might be limiting since we want to support remote data that is symlinked in the future.

One solution would be to create a `PortableCode`, but we established in meetings that this might be too limiting as we sometimes want to only use a single script without the folder.

We the approach to define the bash command and add the script as argument. This is also limiting because one cannot support other scripts than bash. A different solution has to be found.


src/sirocco/workgraph.py
```diff
@@ -221,24 +221,17 @@ def create_task_node(self, task: core.Task):
     def _create_shell_task_node(self, task: core.ShellTask):
         label = self.get_aiida_label_from_graph_item(task)
         # Split command line between command and arguments (this is required by aiida internals)
+
```
agoscinski, line 224:
```suggestion
```

PR #98, opened 2025-01-20
Support python 3.10

agoscinski:
Since python 3.9 will be deprecated this year, I would not bother supporting it, especially because the type hint improvements with 3.10 are much more readable.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,7 +14,7 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
-        python-version: ['3.10', '3.13']
+        python-version: ['3.10', '3.11', '3.12', '3.13']
         aiida-version: ['stable']
 
     steps:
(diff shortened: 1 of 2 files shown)


GeigerJ2 (approved):

.github/workflows/ci.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
-        python-version: ['3.12']
+        python-version: ['3.10', '3.13']
```
DropD, line 17: is this treated as a range?

agoscinski: We can add it, I did not want to add computation time for very because I did not know what the free limit of Github actions was but with 2000 minutes we should be fine
https://docs.github.com/en/billing/managing-billing-for-your-products/managing-billing-for-github-actions/about-billing-for-github-actions

agoscinski:
```suggestion
        python-version: ['3.10', '3.11', '3.12', '3.13']
```

DropD (approved):

PR #100, opened 2025-01-21
Move the canonical class to parsing and core modules

agoscinski:
There is a bug the `Task.plugin_classes` are not yet initialized. Maybe it is because of the constructor approach I take. Need to investigate

### DIFF
--- a/src/sirocco/core/workflow.py
+++ b/src/sirocco/core/workflow.py
@@ -1,12 +1,11 @@
 from __future__ import annotations
 
-from itertools import product
+from itertools import chain, product
 from typing import TYPE_CHECKING, Self
 
 from sirocco.core.graph_items import Cycle, Data, Store, Task
 from sirocco.parsing._yaml_data_models import (
-    CanonicalWorkflow,
-    load_workflow_config,
+    ConfigWorkflow,
 )
 
 if TYPE_CHECKING:
@@ -14,47 +13,67 @@ if TYPE_CHECKING:
     from datetime import datetime
     from pathlib import Path
 
-    from sirocco.parsing._yaml_data_models import ConfigCycle
+    from sirocco.parsing._yaml_data_models import (
+        ConfigAvailableData,
+        ConfigCycle,
+        ConfigData,
+        ConfigGeneratedData,
+        ConfigTask,
+    )
 
 
 class Workflow:
     """Internal representation of a workflow"""
 
-    def __init__(self, workflow_config: CanonicalWorkflow) -> None:
-        self.name: str = workflow_config.name
-        self.config_rootdir: Path = workflow_config.rootdir
+    def __init__(
+        self,
+        name: str,
+        config_rootdir: Path,
+        cycles: list[ConfigCycle],
+        tasks: list[ConfigTask],
+        data: ConfigData,
+        parameters: dict[str, list],
+    ) -> None:
+        self.name: str = name
+        self.config_rootdir: Path = config_rootdir
+
         self.tasks: Store = Store()
         self.data: Store = Store()
         self.cycles: Store = Store()
 
+        data_dict: dict[str, ConfigAvailableData | ConfigGeneratedData] = {
+            data.name: data for data in chain(data.available, data.generated)
+        }
+        task_dict: dict[str, ConfigTask] = {task.name: task for task in tasks}
+
         # Function to iterate over date and parameter combinations
         def iter_coordinates(param_refs: list, date: datetime | None = None) -> Iterator[dict]:
-            space = ({} if date is None else {"date": [date]}) | {k: workflow_config.parameters[k] for k in param_refs}
+            space = ({} if date is None else {"date": [date]}) | {k: parameters[k] for k in param_refs}
             yield from (dict(zip(space.keys(), x, strict=False)) for x in product(*space.values()))
 
         # 1 - create availalbe data nodes
-        for data_config in workflow_config.data.available:
+        for data_config in data.available:
             for coordinates in iter_coordinates(param_refs=data_config.parameters, date=None):
                 self.data.add(Data.from_config(config=data_config, coordinates=coordinates))
 
         # 2 - create output data nodes
-        for cycle_config in workflow_config.cycles:
+        for cycle_config in cycles:
             for date in self.cycle_dates(cycle_config):
                 for task_ref in cycle_config.tasks:
                     for data_ref in task_ref.outputs:
                         data_name = data_ref.name
-                        data_config = workflow_config.data_dict[data_name]
+                        data_config = data_dict[data_name]
                         for coordinates in iter_coordinates(param_refs=data_config.parameters, date=date):
                             self.data.add(Data.from_config(config=data_config, coordinates=coordinates))
 
         # 3 - create cycles and tasks
-        for cycle_config in workflow_config.cycles:
+        for cycle_config in cycles:
             cycle_name = cycle_config.name
             for date in self.cycle_dates(cycle_config):
                 cycle_tasks = []
                 for task_graph_spec in cycle_config.tasks:
                     task_name = task_graph_spec.name
-                    task_config = workflow_config.task_dict[task_name]
+                    task_config = task_dict[task_name]
 
                     for coordinates in iter_coordinates(param_refs=task_config.parameters, date=date):
                         task = Task.from_config(
@@ -88,5 +107,21 @@ class Workflow:
                 yield date
 
     @classmethod
-    def from_yaml(cls: type[Self], config_path: str) -> Self:
-        return cls(load_workflow_config(config_path))
+    def from_config_file(cls: type[Self], config_path: str) -> Self:
+        """
+        Loads a python representation of a workflow config file.
+
+        :param config_path: the string to the config yaml file containing the workflow definition
+        """
+        return cls.from_config_workflow(ConfigWorkflow.from_config_file(config_path))
+
+    @classmethod
+    def from_config_workflow(cls: type[Self], config_workflow: ConfigWorkflow) -> Workflow:
+        return cls(
+            name=config_workflow.name,
+            config_rootdir=config_workflow.rootdir,
+            cycles=config_workflow.cycles,
+            tasks=config_workflow.tasks,
+            data=config_workflow.data,
+            parameters=config_workflow.parameters,
+        )
--- a/src/sirocco/parsing/_yaml_data_models.py
+++ b/src/sirocco/parsing/_yaml_data_models.py
@@ -13,8 +13,8 @@ from typing import Annotated, Any, ClassVar, Literal, Self
 from isoduration import parse_duration
 from isoduration.types import Duration  # pydantic needs type # noqa: TCH002
 from pydantic import (
-    AfterValidator,
     BaseModel,
+    BeforeValidator,
     ConfigDict,
     Discriminator,
     Field,
@@ -680,8 +680,8 @@ class ConfigWorkflow(BaseModel):
 
     rootdir: Path
     name: str
-    cycles: Annotated[list[ConfigCycle], AfterValidator(list_not_empty)]
-    tasks: Annotated[list[ConfigTask], AfterValidator(list_not_empty)]
+    cycles: Annotated[list[ConfigCycle], BeforeValidator(list_not_empty)]
+    tasks: Annotated[list[ConfigTask], BeforeValidator(list_not_empty)]
     data: ConfigData
     parameters: dict[str, list] = {}
 
--- a/src/sirocco/parsing/_yaml_data_models.py
+++ b/src/sirocco/parsing/_yaml_data_models.py
@@ -643,7 +643,7 @@ class ConfigWorkflow(BaseModel):
             ...           tasks:
             ...             - task_a:
             ...     tasks:
-            ...       - task_b:
+            ...       - task_a:
             ...           plugin: shell
             ...     data:
             ...       available:
@@ -664,7 +664,7 @@ class ConfigWorkflow(BaseModel):
             ...     name="minimal",
             ...     rootdir=Path("/location/of/config/file"),
             ...     cycles=[ConfigCycle(minimal_cycle={"tasks": [ConfigCycleTask(task_a={})]})],
-            ...     tasks=[ConfigShellTask(task_b={"plugin": "shell"})],
+            ...     tasks=[ConfigShellTask(task_a={"plugin": "shell"})],
             ...     data=ConfigData(
             ...         available=[
             ...             ConfigAvailableData(name="foo", type=DataType.FILE, src="foo.txt")
--- a/src/sirocco/parsing/_yaml_data_models.py
+++ b/src/sirocco/parsing/_yaml_data_models.py
@@ -8,7 +8,7 @@ from dataclasses import dataclass, field
 from datetime import datetime
 from io import StringIO
 from pathlib import Path
-from typing import Annotated, Any, ClassVar, Literal
+from typing import Annotated, Any, ClassVar, Literal, Self
 
 from isoduration import parse_duration
 from isoduration.types import Duration  # pydantic needs type # noqa: TCH002
@@ -26,6 +26,15 @@ from pydantic import (
 from ruamel.yaml import YAML
 
 from sirocco.parsing._utils import TimeUtils
+
+ITEM_T = typing.TypeVar("ITEM_T")
+
+
+def list_not_empty(value: list[ITEM_T]) -> list[ITEM_T]:
+    if len(value) < 1:
+        msg = "At least one element is required."
+        raise ValueError(msg)
+    return value
 
 
 class _NamedBaseModel(BaseModel):
@@ -625,8 +634,10 @@ class ConfigWorkflow(BaseModel):
         minimal yaml to generate:
 
             >>> import textwrap
-            >>> config = textwrap.dedent(
+            >>> content = textwrap.dedent(
             ...     '''
+            ...     name: minimal
+            ...     rootdir: /location/of/config/file
             ...     cycles:
             ...       - minimal_cycle:
             ...           tasks:
@@ -637,25 +648,40 @@ class ConfigWorkflow(BaseModel):
             ...     data:
             ...       available:
             ...         - foo:
-            ...             type: "file"
-            ...             src: "foo.txt"
+            ...             type: file
+            ...             src: foo.txt
             ...       generated:
             ...         - bar:
-            ...             type: "file"
-            ...             src: some_task_output
+            ...             type: dir
+            ...             src: bar
             ...     '''
             ... )
-            >>> wf = validate_yaml_content(ConfigWorkflow, config)
+            >>> wf = validate_yaml_content(ConfigWorkflow, content)
 
         minimum programmatically created instance
 
-            >>> empty_wf = ConfigWorkflow(cycles=[], tasks=[], data={})
+            >>> wf = ConfigWorkflow(
+            ...     name="minimal",
+            ...     rootdir=Path("/location/of/config/file"),
+            ...     cycles=[ConfigCycle(minimal_cycle={"tasks": [ConfigCycleTask(task_a={})]})],
+            ...     tasks=[ConfigShellTask(task_b={"plugin": "shell"})],
+            ...     data=ConfigData(
+            ...         available=[
+            ...             ConfigAvailableData(name="foo", type=DataType.FILE, src="foo.txt")
+            ...         ],
+            ...         generated=[
+            ...             ConfigGeneratedData(name="bar", type=DataType.DIR, src="bar")
+            ...         ],
+            ...     ),
+            ...     parameters={},
+            ... )
 
     """
 
-    name: str | None = None
-    cycles: list[ConfigCycle]
-    tasks: list[ConfigTask]
+    rootdir: Path
+    name: str
+    cycles: Annotated[list[ConfigCycle], AfterValidator(list_not_empty)]
+    tasks: Annotated[list[ConfigTask], AfterValidator(list_not_empty)]
     data: ConfigData
     parameters: dict[str, list] = {}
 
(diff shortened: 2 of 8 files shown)


tests/unit_tests/parsing/test_yaml_data_models.py
```diff
@@ -2,20 +2,21 @@
 import textwrap
 
 from sirocco.parsing import _yaml_data_models as models
+from sirocco.core import Workflow
 
 
-def test_workflow_canonicalization():
+def test_workflow_():
```
DropD, line 8: A test for `Workflow.from_config_workflow` already exists. If this now tests something in addition to the existing one, move it to `test_workflow` (and deduplicate). Else, remove the module for now.

src/sirocco/parsing/_yaml_data_models.py
```diff
@@ -541,11 +548,12 @@ class ConfigWorkflow(BaseModel):
 
     """
 
+    rootdir: None | Path = None
```
DropD, line 551: We took `rootdir` out of here, because it can not be supplied inside the yaml file, based on Matthieu's comment on #89 . What is the reasoning for reintroducing it here?

src/sirocco/parsing/_yaml_data_models.py
```diff
@@ -570,67 +578,21 @@ def check_parameters(self) -> ConfigWorkflow:
...
-    data: ConfigData
-    parameters: dict[str, list[Any]]
-
-    @property
-    def data_dict(self) -> dict[str, ConfigAvailableData | ConfigGeneratedData]:
-        return {data.name: data for data in itertools.chain(self.data.available, self.data.generated)}
-
-    @property
-    def task_dict(self) -> dict[str, ConfigTask]:
-        return {task.name: task for task in self.tasks}
```
DropD, line 598: Do you not find the property approach more elegant than duplicating this information manually? If you are worried about performance, it can always be turned into a cached property, once we have evidence that it is actually problematic.

src/sirocco/parsing/_yaml_data_models.py
```diff
@@ -541,11 +548,12 @@ class ConfigWorkflow(BaseModel):
 
     """
 
+    rootdir: None | Path = None
     name: str | None = None
-    cycles: list[ConfigCycle]
-    tasks: list[ConfigTask]
+    cycles: Annotated[list[ConfigCycle], AfterValidator(list_not_empty)]
+    tasks: Annotated[list[ConfigTask], AfterValidator(list_not_empty)]
```
DropD, line 554: If we can add the validators here, then the `CanonicalWorkflow` class only adds the rootdir. I agree that then, there is a more elegant solution that duplicates less code and is orthogonal to the changes to `workflow.Workflow` in this PR.

src/sirocco/core/workflow.py
```diff
@@ -1,60 +1,77 @@
...
+
+        self.name: str = name
+        self.config_rootdir: Path = rootdir # TODO rename to rootdir since it does not make sense with this constructor
+
         self.tasks: Store = Store()
         self.data: Store = Store()
         self.cycles: Store = Store()
 
+        data_dict: dict[str, ConfigAvailableData | ConfigGeneratedData] = {data.name: data for data in chain(data.available, data.generated)}
+        task_dict: dict[str, ConfigTask] = {task.name: task for task in tasks}
```
DropD, line 45: The properties could also be here (see the other comment about properties).

src/sirocco/core/workflow.py
DropD, line 1: These changes strike me as orthogonal to the goal of canonicalization. They could be isolated and put into their own PR.

src/sirocco/core/workflow.py
DropD: or, for performance reasons, they could store the output of the properties.

agoscinski: I think these are just temporaries that are needed for the creation of the Workflow. One can add them as properties but they are nowhere else in the code used

src/sirocco/parsing/_yaml_data_models.py
agoscinski: It can be supplied by using the `from_config_file` constructor, it seems to me also reasonable that `ConfigWorkflow` is aware of its location. It is not perfect since you are right that the information is not supplied in the yaml file. The alternative, however, to make the `core.Workflow` aware of how the `ConfigWorkflow` should be used for parsing seems to me less ideal. So I think it is not the perfect but the alternative seems worse to me

tests/unit_tests/parsing/test_yaml_data_models.py
agoscinski: ok! will check this

src/sirocco/parsing/_yaml_data_models.py
agoscinski: I just did not see where it is needed except for the conversion to `core.Workflow`.

src/sirocco/core/workflow.py
DropD: I meant isolated from other changes, like undoing the shift to properties instead of duplicating information, moving the validators to the ConfigWorkflow and replacing CanonicalWorkflow with some other way of storing ConfigWorkflow + rootdir.

src/sirocco/core/workflow.py
DropD: Then properties are the way to go.

src/sirocco/parsing/_yaml_data_models.py
DropD: In what way was `core.Workflow` aware of any of this in #89? Does it not provide exactly that alternative without the drawbacks?

src/sirocco/parsing/_yaml_data_models.py
DropD: I would suggest replacing `CanonicalWorkflow` with a dataclass / pydantic model / namedtuple (whatever), which puts a type name to `ConfigWorkflow` + `rootdir`.  `workflow.Workflow` should then only deal with that type.

src/sirocco/parsing/_yaml_data_models.py
agoscinski: You are right, this logic was in `_yaml_data_models` so `core.Workflow` was not aware of the parsing logic, I remembered it wrongly. Before the canonicalization the `rootdir` was updated in the  `_yaml_data_models.load_workflow_config` function. Then this was moved to the construction of the `CanonicalWorkflow`. In this PR we move it to the `ConfigWorkflow` into the new `from_config_file` constructor so it is handled within the class which is more logic than adding it an util function as it was before the canonicalization. I think this an implementation imitating the benefits of the `CanonicalWorkflow` but moving it to the `ConfigWorkflow`. One can use `rootdir` is an indicator if the workflow was created from a file or not.

src/sirocco/core/workflow.py
agoscinski: This I don't fully understand, why introduce a property for something that is computed once? You expose it to the API of the class. Do you want to capsulate the logic to test it? This I could understand, but then I would make them private properties.

src/sirocco/core/workflow.py
agoscinski: I can split it up maybe in multiple commits?

src/sirocco/core/workflow.py
agoscinski: Hm.. no as property it also does not make sense as it is created from the `tasks: ConfigTask` that are only provided during initialization in `core.Workflow`, so accessing it afterwards does not make so much sense. If you want to test it I would propose to imitate the `build_internal_dicts` function in `core.Workflow`
```python
    def _build_data_dict(self, data: ConfigData) -> dict[str, ConfigAvailableData | ConfigGeneratedData]:
        return data_dict = {data.name: data for data in data.available} | {
            data.name: data for data in data.generated
        }
```
same for task dict

tests/unit_tests/parsing/test_yaml_data_models.py
agoscinski: ok done!

src/sirocco/core/workflow.py
agoscinski: I did it now like this, what do you think?

src/sirocco/core/workflow.py
```diff
@@ -1,60 +1,80 @@
...
-        self.name: str = workflow_config.name
-        self.config_rootdir: Path = workflow_config.rootdir
+    def __init__(self,
+                 name: str,
+                 rootdir: Path,
+                 cycles: list[ConfigCycle],
+                 tasks: list[ConfigTask],
+                 data: ConfigData,
+                 parameters: dict[str, list], # TODO type should be more specified but not in this PR
+                 ) -> None:
```
DropD, line 35: This signature change is valuable, especially together with the `from_config_workflow` classmethod.

src/sirocco/core/workflow.py
```diff
@@ -1,60 +1,80 @@
...
+
+        self.name: str = name
+        if not isinstance(rootdir, Path):
+            msg = f"Need to provide path for rootdir but got {type(rootdir)}"
+            raise TypeError(msg)
```
DropD, line 40: Moving this logic here is what we agreed not to do, as it is part of dealing with user input.

src/sirocco/core/workflow.py
```diff
@@ -80,13 +100,38 @@ def iter_coordinates(param_refs: list, date: datetime | None = None) -> Iterator
         for task in self.tasks:
             task.link_wait_on_tasks(self.tasks)
 
+    @staticmethod
+    def _build_data_dict(data: ConfigData) -> dict[str, ConfigAvailableData | ConfigGeneratedData]:
+        return {data.name: data for data in chain(data.available, data.generated)}
+    @staticmethod
+    def _build_tasks_dict(tasks: list[ConfigTask]) -> dict[str, ConfigTask]:
+        return {task.name: task for task in tasks}
```
DropD, line 108: If these were instance methods, there would be no need to pass `data` / `tasks` in there. Since they are prefixed with `_` you are declaring they should not be used outside an instance (where passing the information in would make sense). What is the design rationale for this contradiction?

src/sirocco/parsing/_yaml_data_models.py
DropD: Please explain in which way moving `from_config_file` from the module to the class makes it "more logical"?

src/sirocco/parsing/_yaml_data_models.py
```diff
@@ -570,67 +578,19 @@ def check_parameters(self) -> ConfigWorkflow:
...
-
-    rootdir = config_path.resolve().parent
-
-    return canonicalize_workflow(config_workflow=parsed_workflow, rootdir=rootdir)
-    # return parsed_workflow
+    @classmethod
+    def from_config_file(cls, config_filename: str):
+        # TODO maybe not best design, since we have multiple entry points, but
+        #      on the other hand we don't have rootdir ambiguity,I think this can be
+        #      better integrated with pydantic
```
DropD, line 585: In which way is it true that "we don't have rootdir ambiguity"? `.rootdir` is still type hinted as `Path | None` and a check in `workflow.Workflow.__init__` is still required, is there another definition of "ambiguity"?

src/sirocco/parsing/_yaml_data_models.py
DropD: If we do this, this PR should add versions of each test that uses `ConfigWorkflow` with `rootdir=None` and check the appropriate outcome.

src/sirocco/parsing/_yaml_data_models.py
```diff
@@ -541,11 +548,12 @@ class ConfigWorkflow(BaseModel):
 
     """
 
+    rootdir: None | Path = None
     name: str | None = None
```
DropD, line 552: Since now ambiguity in the type of `name` is reintroduced, all tests are required to have a version for `name=None`

src/sirocco/parsing/_yaml_data_models.py
```diff
@@ -23,6 +23,15 @@
 
 from sirocco.parsing._utils import TimeUtils
 
+ITEM_T = typing.TypeVar("ITEM_T")
```
agoscinski, line 26: If the type is not specified why bother specifying it? Isn't list[ITEM_T] similar useful for type checking as list at the moment? Or will this type more explicitly specified in the future?

src/sirocco/core/workflow.py
leclairm: I second that

src/sirocco/parsing/_yaml_data_models.py
leclairm: I presume that if it is used only once it doesn't make sense but when used for several input types or both input and output, it indicates those must be the same.

src/sirocco/parsing/_yaml_data_models.py
leclairm: > If we can add the validators here, then the `CanonicalWorkflow` class only adds the rootdir. I agree that then, there is a more elegant solution that duplicates less code and is orthogonal to the changes to `workflow.Workflow` in this PR.

 `CanonicalWorkflow` also makes `name` non-optional

src/sirocco/core/workflow.py
agoscinski: Okay removed, see latest commit

src/sirocco/parsing/_yaml_data_models.py
agoscinski: > Please explain in which way moving from_config_file from the module to the class makes it "more logical"? 

Since it is essentially a method that constructs an instance of the class.

> If we do this, this PR should add versions of each test that uses ConfigWorkflow with rootdir=None and check the appropriate outcome.

Added tests

src/sirocco/parsing/_yaml_data_models.py
agoscinski: I used now a reasonable default argument for `rootdir` that I would have been favor for in a future PR, but since it simplifies the handling of the `rootdir` with both constructors, I introduced it here.

src/sirocco/parsing/_yaml_data_models.py
agoscinski: Yes but we introduce it without doing this.

src/sirocco/parsing/_yaml_data_models.py
agoscinski: If we move `rootdir` outside of the ConfigWorkflow I think it would become problematic with setting `rootdir` inside the yaml file

agoscinski: I realized that the user input in the file was not be respected so I had to to open up the parsing. I realized that the parsing of pydantic_yaml is actually not using pydantic v2 so I would remove it in a future PR. As a consequence the default argument for `rootdir` was not required anymore to get a technical simpler solution, so I removed it to reintroduce this in some later PR.

src/sirocco/parsing/_yaml_data_models.py
DropD: Here is how to read `def list_not_empty(value: list[ITEM_T]) -> list[ITEM_T]:`:

- `list_not_empty` is a function `list[T] -> list[T]`, therefore
- it is allowed to take `[1, 2, 3]` and return `[4, 5, 6, 7, 8]
- it is not allowed to take `[1, 2, 3]` and return `["a", "b", "c"]`

The guarantee is that the mix of types in the input list is the same as in the output list

Here is how to read `def list_not_empty(value: list) -> list:`:

- `list_not_empty` is a function `list[Any] -> list[Any]`, therefore
- it is allowed to take `[1, 2, 3]` and return `["1", "2", "3"]`
- it is even allowed to take `[1, 2, 3]` and return `[("a", "b"), 4.5, None, [[[1], 2], 3]]`

The guarantee on the relationship of item types in the input and output list is absolutely none.

This is because `Any` is the union of all types.

DropD: This function is generic, but it is core to it's design that it does not change the input list. The closest we can do to letting type checkers know that is this. Now type checkers don't have to complain if we use `cycles` as a list of `Cycle`s after running this function on it. Otherwise it would have to assume that it can now contain anything.

DropD: > Isn't list[ITEM_T] similar useful for type checking as list at the moment?
In conclusion, no. It is a lot more useful, which will become super obvious when switching on type checking project wide in CI.

agoscinski: okay I get the difference to just use `list[Any]` thanks for the clarification

src/sirocco/parsing/_yaml_data_models.py
DropD: I had sort of assumed that there was a design decision to not include `rootdir` in the yaml config. If we can (and also make it and `name` mandatory) then `CanonicalWorkflow` can be removed without replacement.

src/sirocco/core/workflow.py
```diff
@@ -88,5 +107,21 @@ def cycle_dates(cycle_config: ConfigCycle) -> Iterator[datetime]:
...
+        Loads a python representation of a workflow config file.
+
+        :param config_filename: the string to the config yaml file containing the workflow definition
+        """
+        return cls.from_config_workflow(ConfigWorkflow.from_config_file(config_filename))
+
+    @classmethod
+    def from_config_workflow(cls: type[Self], config_workflow: ConfigWorkflow) -> Workflow:
+        return cls(
+            name=config_workflow.name,
```
DropD, line 121: This will fail mypy checks, because `config_workflow.name` can be `None`. That was the main job of `CanonicalWorkflow` to avoid.

src/sirocco/core/workflow.py
```diff
@@ -88,5 +107,21 @@ def cycle_dates(cycle_config: ConfigCycle) -> Iterator[datetime]:
...
+            rootdir=config_workflow.rootdir,
```
DropD, line 122: This will fail mypy checks, because `ConfigWorkflow.rootdir` can be None. That was the main job of CanonicalWorkflow to avoid.

src/sirocco/core/workflow.py
DropD: This PR needs to catch that now. In general, please run `mypy` manually and check that it isn't a problem anywhere. This is the bar for any PR that wants to replace `CanonicalWorkflow` (You can run mypy there and compare).

src/sirocco/core/workflow.py
DropD: See [the other comment about this](https://github.com/C2SM/Sirocco/pull/100/files#r1930575335).

src/sirocco/parsing/_yaml_data_models.py
agoscinski: > I had sort of assumed that there was a design decision to not include rootdir in the yaml config

I am not sure. I think the name `rootdir` indicates that it should not be set by the user, but right now it acts more like a working directory which I would argue should be settable by the user. One could introduce a new variable that makes this difference apparent. On the other hand, the only the usage of `rootdir` is as a working directory, ergo one could replace it. What do you think @DropD @leclairm

src/sirocco/core/workflow.py
agoscinski: `rootdir` and `name` cannot be None anymore

src/sirocco/parsing/_yaml_data_models.py
agoscinski: `rootdir` and `name` cannot be None anymore

src/sirocco/parsing/_yaml_data_models.py
leclairm: `rootdir` is the directory where we find the full configuration, which consists not only of the yaml file but also all the other ones like ICON namelists, user scripts for shell tasks, etc... Subsequently, this is also the root path used to resolve the relative paths provided in the yaml file for these other config files. So it has to simply be the parent directory of the yaml file, nothing else, and must not be set by the user in the yaml file.

Regarding its name, `rootdir` seems logical as it is originally a property of `CanonicalWorkflow` being itself derived from `ConfigWorkflow`. Then it becomes `Workflow.config_rootdir` which still makes sense. The only name we could change here for consistency is `CanonicalWorkflow` -> `CanonicalConfigWorkflow` (if the canonical class is kept, I still need to read through the PR again to understand what it enables)

leclairm (changes requested):

leclairm:
I'm fine with the PR, once we've handled the `ConfigWorkflow.rootdir` issue (cannot be set by the user).

src/sirocco/parsing/_yaml_data_models.py
agoscinski: > rootdir is the directory where we find the full configuration, which consists not only of the yaml file but also all the other ones like ICON namelists, user scripts for shell tasks, etc... Subsequently, this is also the root path used to resolve the relative paths provided in the yaml file for these other config files. So it has to simply be the parent directory of the yaml file, nothing else, and must not be set by the user in the yaml file.

But here again you argue that there is must be `rootdir`, then we use it as working directory. I argue there must be working directory, and if this working directory is not specified we use the `rootdir` as it. I don't see why we should not give the user the possibility to specify where the data for the workflow is so the user has more flexibility in arranging his/her folder structure.

src/sirocco/core/workflow.py
agoscinski: `name` cannot be anymore None, by separating the yaml parsing and validation one can set it in the `from_config_file` constructor. The only reason why we allowed `None` was because before we could not determine if `name` is specified in the given file.

src/sirocco/parsing/_yaml_data_models.py
```diff
@@ -570,67 +596,21 @@ def check_parameters(self) -> ConfigWorkflow:
...
+
+        config_path = Path(config_filename)
+        content = config_path.read_text()
+        reader = YAML(typ="safe", pure=True)
+        object_ = reader.load(StringIO(content))
+        # If name was not specified, then we use filename without file extension
+        if "name" not in object_:
+            object_["name"] = config_path.stem
+        if "rootdir" not in object_:
+            object_["rootdir"] = config_path.resolve().parent
```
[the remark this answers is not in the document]
agoscinski: I was first advocating for being able to manipulate the working directory, but when I thought this through I came to the conclusion that this not important (see issue https://github.com/C2SM/Sirocco/issues/111).

src/sirocco/parsing/_yaml_data_models.py
```diff
@@ -533,17 +546,30 @@ class ConfigWorkflow(BaseModel):
...
 
     """
 
-    name: str | None = None
-    cycles: list[ConfigCycle]
-    tasks: list[ConfigTask]
+    rootdir: Path
+    name: str
+    cycles: Annotated[list[ConfigCycle], AfterValidator(list_not_empty)]
+    tasks: Annotated[list[ConfigTask], AfterValidator(list_not_empty)]
```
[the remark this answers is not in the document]
agoscinski: I just replicated the existing check. In this case it does not make a real difference. If I am not wrong, it just specifies if it happens before the cycles are converted from list of dictionaries to a list of ConfigCycles or after. Doing it after or before should not change it logically. But yes we could change it to `BeforeValidator`.

src/sirocco/core/workflow.py
```diff
@@ -1,60 +1,79 @@
...
 class Workflow:
     """Internal representation of a workflow"""
 
-    def __init__(self, workflow_config: CanonicalWorkflow) -> None:
-        self.name: str = workflow_config.name
-        self.config_rootdir: Path = workflow_config.rootdir
+    def __init__(
+        self,
+        name: str,
+        rootdir: Path,
```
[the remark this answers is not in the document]
agoscinski: I think this will be changed once rebased. Waiting for PR #110 being merged

src/sirocco/parsing/_yaml_data_models.py
```diff
@@ -570,67 +596,21 @@ def check_parameters(self) -> ConfigWorkflow:
...
-    # If name was not specified, then we use filename without file extension
-    if parsed_workflow.name is None:
-        parsed_workflow.name = config_path.stem
-
-    rootdir = config_path.resolve().parent
-
-    return canonicalize_workflow(config_workflow=parsed_workflow, rootdir=rootdir)
-    # return parsed_workflow
+    @classmethod
+    def from_config_file(cls, config_filename: str):
```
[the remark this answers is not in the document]
agoscinski: depends, path could also mean that you need to specify a `pathlib.Path`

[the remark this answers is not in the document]
agoscinski: ah okay, yes in this case we actually specify a path not a filename, will do it

src/sirocco/core/workflow.py
```diff
@@ -88,5 +107,21 @@ def cycle_dates(cycle_config: ConfigCycle) -> Iterator[datetime]:
                 yield date
 
     @classmethod
-    def from_yaml(cls: type[Self], config_path: str) -> Self:
-        return cls(load_workflow_config(config_path))
+    def from_yaml(cls: type[Self], config_filename: str) -> Self:
```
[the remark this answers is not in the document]
agoscinski: yes I wanted to change this in another PR to keep it separate, but one could also do it in this PR and make another commit

src/sirocco/parsing/_yaml_data_models.py
agoscinski: We decided in a meeting to keep `rootdir` as a name because `working_directory` implies too much that the calculation happens there

src/sirocco/core/workflow.py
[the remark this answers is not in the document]
agoscinski:
```suggestion
    def from_ config_file(cls: type[Self], config_file_path: str) -> Self:
```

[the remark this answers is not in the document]
agoscinski: changed now in this PR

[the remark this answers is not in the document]
agoscinski: I used `config_path` since this name was used in the visualizer

src/sirocco/parsing/_yaml_data_models.py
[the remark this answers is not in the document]
agoscinski: changed, I used config_path since this name was used in the visualizer

tests/unit_tests/parsing/test_yaml_data_models.py
```diff
@@ -63,6 +46,38 @@ def test_load_workflow_config(tmp_path):
...
+
+    with pytest.raises(ValidationError, match=r".*1 validation error for ConfigWorkflow\nrootdir.*"):
+        models.ConfigWorkflow(
+            name=minimal_config.name,
+            rootdir=None,
+            cycles=minimal_config.cycles,
+            tasks=minimal_config.tasks,
+            data=minimal_config.data,
+            parameters=minimal_config.parameters,
+        )
```
agoscinski, line 83: TODO remove

DropD (approved):

tests/unit_tests/parsing/test_yaml_data_models.py
agoscinski, line 83:
```suggestion
```

leclairm (approved):

PR #101, opened 2025-01-21
Improve ConfigData & related

DropD:
- Add Tests
- Rewrite `_NamedBaseModel` to allow alternative creation of models with intuitive kwargs from python
- Make `.type` and `.src` non-optional in `ConfigBaseDataSpecs`
- Change `ConfigBaseDataSpecse.type` type to an enum (replaces string-based validator while improving legibility)
- In combination no need to add a "canonical" version.


src/sirocco/parsing/_yaml_data_models.py
```diff
@@ -478,7 +521,36 @@ def invalid_field(cls, value: str | None) -> str | None:
...
+            >>> snippet = textwrap.dedent(
+            ...     '''
+            ...     available:
+            ...       - foo:
+            ...           type: "file"
+            ...           src: "foo.txt"
+            ...     generated:
+            ...       - name: "bar"
+            ...         type: "file"
+            ...         src: "bar.txt"
```
leclairm, line 542: We should discuss if it makes sense to document both syntax. It's no pb internally in the doctest. For the user documentation I think 2 syntax would bring confusion. Maybe early users could give feedback on this.

src/sirocco/parsing/_yaml_data_models.py
```diff
@@ -25,46 +26,64 @@
 
 
 class _NamedBaseModel(BaseModel):
-    """Base class for all classes with a key that specifies their name.
-
-    For example:
-
-    .. yaml
+    """
+    Noninvasive base model for reading names from yaml keys *or* attributes.
```
agoscinski, line 30: What do you mean with "noninvasive"? Maybe this can be made more clear

src/sirocco/parsing/_yaml_data_models.py
```diff
@@ -25,46 +26,64 @@
...
+                            data = value | {"name": key}
+                        case None:
+                            data = {"name": key}
+                        case _:
+                            msg = f"{cls.__name__} may only be used for named objects, not values (got {data})."
+                            raise TypeError(msg)
+                case _:
+                    msg = f"{cls.__name__} requires name to be a str (got {key})."
+                    raise TypeError(msg)
+        return data
```
agoscinski, line 86: Does this make sense to be a classmethod? I typically see classmethods used only for creating an instance the class (so this function would `dict` -> `_NamedBaseModel`). Here we create a dict object. Don't you think a `staticmethod` is more transparently communicating that not an instance of the same class is created?

src/sirocco/parsing/_yaml_data_models.py
agoscinski: This is a docstring test, so we run this code when running `hatch test` which is very useful to verify that it works. I think the end user is not supposed to read the in-code docstrings but only the docpage so there we can only document one use case.

src/sirocco/parsing/_yaml_data_models.py
DropD: True, I will adjust the docstring. What I mean is that it doesn't disable the standard Python API.

src/sirocco/parsing/_yaml_data_models.py
DropD: My reasons:
- Model validators in pydantic docs are classmethods (this is only a separate method to allow additional usage).
- It allows access to `cls.__name__`, which will report the name of the subclass. This is more helpful and less confusing than always reporting the name of the base class.
- This class method is not designed to deal with arbitrary dicts. Semantically, it is not intended to have meaning outside this class. Therefore it violates the rule "staticmethod could also be standalone function"
- If there is value in it being designed for standalone usage, I would prefer to make it a standalone function and design and test it accordingly.

src/sirocco/parsing/_yaml_data_models.py
DropD: I have changed this to be the intended syntax for both, because:
- We agree that it is a side effect, and if we find a way to remove it, we might.
- It is tested in the base class, both Python API and yaml snippet, for internal documentation already.

src/sirocco/parsing/_yaml_data_models.py
DropD: I actually changed the usage in `get_plugin_from_named_base_model` to use the `ConfigBaseTask.extract_merge_name`, which should cause errors to report the right class name there as well.

agoscinski: ok good points!

src/sirocco/parsing/_yaml_data_models.py
```diff
@@ -25,46 +26,67 @@
...
 class _NamedBaseModel(BaseModel):
-    """Base class for all classes with a key that specifies their name.
-
-    For example:
+    """
+    Base model for reading names from yaml keys *or* attributes.
 
-    .. yaml
+    Reading from attributes is allowed in order to enable the standard
+    constructor usage from Python, as demonstrated in the below examples.
```
agoscinski, line 33: I must say this is quite pedantic but I had some troubles to understand the text: If I understand the [yaml specification](https://yaml.org/spec/1.2.2/#nodes), they dont use the termology "attributes" but "values"

```suggestion
    Reading from values is allowed in order to enable the standard
    constructor usage from Python, as demonstrated in the below examples.
```

src/sirocco/parsing/_yaml_data_models.py
```diff
@@ -25,46 +26,67 @@
 
 
 class _NamedBaseModel(BaseModel):
-    """Base class for all classes with a key that specifies their name.
-
-    For example:
+    """
+    Base model for reading names from yaml keys *or* attributes.
```
agoscinski, line 30: Thank you I think this is more understandable already. I still have a suggestion, I must say this is quite pedantic, but I had some troubles to understand the word "attribute": If I understand the [yaml specification](https://yaml.org/spec/1.2.2/#nodes), they dont use the terminology "attributes" but "values".

```suggestion
    Base model for reading names from yaml keys *or* values.
```

src/sirocco/parsing/_yaml_data_models.py
```diff
@@ -489,7 +564,7 @@ def get_plugin_from_named_base_model(
 ) -> str:
     if isinstance(data, (ConfigRootTask, ConfigShellTask, ConfigIconTask)):
         return data.plugin
-    name_and_specs = _NamedBaseModel.merge_name_and_specs(data)
+    name_and_specs = ConfigBaseTask.extract_merge_name(data)
```
agoscinski, line 567: Since this is the `Data` class maybe we use the `ConfigBaseData`
```suggestion
    name_and_specs = ConfigBaseData.extract_merge_name(data)
```
I don't think this is important

agoscinski (approved):
Just had two minor suggestions . I think they are really marginal so I already approve. Thanks for the work!

leclairm (approved):

src/sirocco/parsing/_yaml_data_models.py
leclairm: The argument of `get_plugin_from_named_base_model` actually is a task from which you want to get the `plugin`. The misunderstanding comes from the argument name, `data`, that should rather be `task`

src/sirocco/parsing/_yaml_data_models.py
DropD: I should actually have mentioned "keyword args to the constructor", since that was the real goal.

PR #103, opened 2025-01-22
WIP: By default run tests in tests folder

agoscinski:
Right now `hatch test` runs also files in top level directory, this is a bit annoying since one needs to specify `hatch test -- tests`. This is not really fixing it, I need to investigate this further.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,7 +78,7 @@ path = "src/sirocco/__init__.py"
 extra-dependencies = [
     "ipdb"
 ]
-default-args = []
+default-args = ["tests"]
 extra-args = ["--doctest-modules"]
 
 [[tool.hatch.envs.hatch-test.matrix]]


DropD:
> Right now `hatch test` runs also files in top level directory, this is a bit annoying since one needs to specify `hatch test -- tests`. This is not really fixing it, I need to investigate this further.

I never need to do that. Do you keep uncommitted `test_xyz.py` files in your top-level, which pytest recognizes as tests?

agoscinski:
It runs for some reason a file called `debug.py`. There is also no test function in it. I dont fully understand this because I don't have this problem in other projects. This PR is more like a reminder to look at it. I dont think the change here really is a good solution.

DropD:
> It runs for some reason a file called `debug.py`. There is also no test function in it. I dont fully understand this because I don't have this problem in other projects. This PR is more like a reminder to look at it. I dont think the change here really is a good solution.

It might be scanning the file for doctests. Since the file is not part of the repo I would recommend trying the following in a top-level `conftest.py`, which you can locally git-ignore: https://stackoverflow.com/questions/41358778/can-i-make-the-pytest-doctest-module-ignore-a-file

agoscinski:
> It might be scanning the file for doctests. Since the file is not part of the repo I would recommend trying the following in a top-level conftest.py, which you can locally git-ignore: https://stackoverflow.com/questions/41358778/can-i-make-the-pytest-doctest-module-ignore-a-file

No it does not work, can I add tests as default arg now?

PR #110, opened 2025-01-27
Remove `pydantic-yaml` as dependency

agoscinski:
The amount of functionality we use of this library is extremely minimal. We replace it with the dependency `ruamel.yaml` that parses the yaml file into a python object. This is basically the code we use from the dependency https://github.com/NowanIlfideme/pydantic-yaml/blob/130569dabbe9843771bc6f7d52830e061238d61b/src/pydantic_yaml/_internals/v2.py#L217-L242

Furthermore, the library still uses pydantic API v1 for validating the pydantic model which is deprecated so we can replace it with pydantic v2. In addition this change gives us more flexibility as it separates the parsing of the yaml file and the validation through pydantic.

To have a similar utility function that can be used in the docstring tests we implement `validate_yaml_content`.

### DIFF
--- a/src/sirocco/parsing/_yaml_data_models.py
+++ b/src/sirocco/parsing/_yaml_data_models.py
@@ -6,6 +6,7 @@ import time
 import typing
 from dataclasses import dataclass, field
 from datetime import datetime
+from io import StringIO
 from pathlib import Path
 from typing import Annotated, Any, ClassVar, Literal
 
@@ -18,9 +19,11 @@ from pydantic import (
     Discriminator,
     Field,
     Tag,
+    TypeAdapter,
     field_validator,
     model_validator,
 )
+from ruamel.yaml import YAML
 
 from sirocco.parsing._utils import TimeUtils
 
@@ -41,8 +44,8 @@ class _NamedBaseModel(BaseModel):
         >>> _NamedBaseModel(foo={})
         _NamedBaseModel(name='foo')
 
-        >>> import pydantic_yaml, textwrap
-        >>> pydantic_yaml.parse_yaml_raw_as(
+        >>> import textwrap
+        >>> validate_yaml_content(
         ...     _NamedBaseModel,
         ...     textwrap.dedent('''
         ...     foo:
@@ -50,7 +53,7 @@ class _NamedBaseModel(BaseModel):
         ... )
         _NamedBaseModel(name='foo')
 
-        >>> pydantic_yaml.parse_yaml_raw_as(
+        >>> validate_yaml_content(
         ...     _NamedBaseModel,
         ...     textwrap.dedent('''
         ...     name: foo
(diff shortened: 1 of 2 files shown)


src/sirocco/parsing/_yaml_data_models.py
```diff
@@ -713,3 +711,31 @@ def load_workflow_config(workflow_config: str) -> CanonicalWorkflow:
...
+
+    Raises:
+        pydantic.ValidationError: If the YAML content cannot be validated
+        against the specified class type.
+        ruamel.yaml.YAMLError: If there is an error in parsing the YAML content.
+    """
+    from io import StringIO
+
+    from pydantic import TypeAdapter
+    from ruamel.yaml import YAML
```
leclairm, line 739: Why importing locally to the function? This will be executed each time the function is called.

leclairm:
Other than my small comment I'm fine with the PR. I thought `pydantic-yaml` was developped by the same people as `pydantic` but apparently not. Hence the lag in using v2 I guess.

Also may I ask why `ruamel,yaml`?

src/sirocco/parsing/_yaml_data_models.py
agoscinski: So I kind of imitated the style before. I think @DropD introduced this in a PR. My understanding is that this is only loaded if the function is called (and then cached so rerunning is not an issue). This makes it faster to do a top level import (`import sirocco`) since not all dependencies that are used in the library are immediately loaded. I am not sure about a guideline here. It makes sense to do it for `ruamel` since it will be only needed in a few places and importing it can take some time, but if for `TypeAdapter`and `StringIO` I am not sure if makes such a difference. Maybe @DropD has a comment or correction.

agoscinski:
> Also may I ask why ruamel.yaml?

I just used the library that pydantic-yaml was using. We have now the flexibility to change it but I did not investigate the differences. Since it is used with particular options `YAML(typ="safe", pure=True)`. One could definitely improve this in the future, for example, `pure=True` uses the default standard Python implementation to load the yaml which is more compatible over systems over using the C-extension but is also slower. I would investigate this in the future when we have a more robust CI (including CI on alps) that could catch potential errors when switching to less compatible implementation. I should make an issue for this.

leclairm:
> I should make an issue for this

Thanks for the explanation but don't bother, this is really low priority.

src/sirocco/parsing/_yaml_data_models.py
leclairm: Also, it hides in the function what the module depends on. Curious about @DropD thoughts on that.

GeigerJ2 (approved):
LGTM

agoscinski:
We decided to move the imports to the header for better overview, the advantage of not importing `ruamel` when using the library only with the python API is not really meaningful for us. So to have a better overview is more important

PR #113, opened 2025-01-30
Use uv for environment management and ruff with py310

leclairm:


src/sirocco/core/graph_items.py
```diff
@@ -53,7 +53,7 @@ def from_config(cls, config: ConfigBaseData, coordinates: dict) -> Self:
 
 
 # contains the input data and its potential associated port
-BoundData: TypeAlias = tuple[Data, str | None]
+BoundData: type = tuple[Data, str | None]
```
agoscinski, line 56: Did the type checker complain? This seemed correct to me, we are defining a type for type hinting and not a type as a new class would be. A type would be something like this

```python
@classmethod
def from_something_constructor(cls: type):
    ...
```
I think one can think about TypeAlias as a specific type value (similar as `int` and the value `5` are related, 5 is  an instance of an `int`)

src/sirocco/core/workflow.py
```diff
@@ -30,7 +30,7 @@ def __init__(self, workflow_config: CanonicalWorkflow) -> None:
         # Function to iterate over date and parameter combinations
         def iter_coordinates(param_refs: list, date: datetime | None = None) -> Iterator[dict]:
             space = ({} if date is None else {"date": [date]}) | {k: workflow_config.parameters[k] for k in param_refs}
-            yield from (dict(zip(space.keys(), x)) for x in product(*space.values()))
+            yield from (dict(zip(space.keys(), x, strict=False)) for x in product(*space.values()))
```
agoscinski, line 33: I would not mix it with this PR, or put it in a separate commit and merge as rebase.

agoscinski (changes requested):

pyproject.toml
```diff
@@ -75,6 +75,7 @@ include = [
 path = "src/sirocco/__init__.py"
 
 [tool.hatch.envs.hatch-test]
```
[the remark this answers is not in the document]
leclairm: yes

src/sirocco/core/graph_items.py
leclairm: Thank you for the little course about what a `type` is :stuck_out_tongue_closed_eyes: 
And yes, ruff complained.

src/sirocco/core/workflow.py
leclairm: Same, then ruff complains.

agoscinski: right, I don' know why I did not conclude it on this change^^

agoscinski (approved):
I split the changes for the formatter and uv into two commits and reintroduced the TypeAlias because I think this was correct

PR #117, opened 2025-01-31
infra: Improve testing usability

DropD:
# Improve testing usability

## mark slow tests and deselect them by default (but not on CI)

Mark the test running AiiDA as slow. Change the default test args to include -m "not slow", which can be overridden by passing -m "".

Now to run the fast tests during development we can use the default
```
hatch test
```
Whereas on CI or to manually run everything we use
```
hatch test -m ""
```
which overrides any selection based on marks (consequently selecting everything).

## ignore warnings from libraries and default to treating warnings as errors

This will save us from building up third party warnings from code out of our control in the test output until the actual test results get pushed out of the console history (tends to happen in python projects over time).
Now we have to decide for every warning we get, whether we can / need to deal with it or should ignore it explicitly in the config.

## cmdline option for regenerating serialized references

The skipped tests which could be manually un-skipped to update references had their logic moved into a config hook for pytest, which checks for a custom cmdline option. We can now:

```
hatch test --reserialize
```

to regenerate everything (can be differenciated in the future by type of data if regenerating everything ever becomes too slow).

## parallelize tests on CI

`--parallel` option was added to `hatch test` on CI (github runners can typically not run many workers, so the speedup is probably small)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -38,7 +38,7 @@ jobs:
       env:
         PYTEST_ADDOPTS: "--durations=0"
       run: |
-        hatch test -m "" --cover # override default exclusion of slow tests with -m ""
+        hatch test -m "" --parallel --cover # override default exclusion of slow tests with -m ""
 
   docs:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -38,7 +38,7 @@ jobs:
       env:
         PYTEST_ADDOPTS: "--durations=0"
       run: |
-        hatch test --cover
+        hatch test -m "" --cover # override default exclusion of slow tests with -m ""
 
   docs:
     runs-on: ubuntu-latest
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -92,6 +92,7 @@ installer = "uv"
 extra-dependencies = [
     "ipdb"
 ]
+default-args = []
 extra-args = ["--doctest-modules", '-m not slow']
 
 [[tool.hatch.envs.hatch-test.matrix]]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,6 +51,13 @@ norecursedirs = "tests/cases"
 markers = [
   "slow: slow integration tests which are not recommended to run locally for normal development"
 ]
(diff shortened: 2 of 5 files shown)


DropD:
now `hatch test` does not run doctests, but `hatch test -v` does, while neither runs the "slow" tests.

leclairm:
> now `hatch test` does not run doctests, but `hatch test -v` does, while neither runs the "slow" tests.

Some observations:
- `hatch test -v` enables the doctests again
-  setting `extra-args = ["--doctest-modules", '-m not slow', '-v']` doesn't help,  doctests are still skipped while the output is verbose.
- Now, still with `-v` in the `extra-args`, running `hatch test -v`, which should make no difference, does enable the doctests again.

So, apparently, appending an extra arg (probably random) re-enables doctests

leclairm:
Also `hatch test --doctest-modules` activates the doctests as expected

leclairm:
Finally found a workaround [there](https://github.com/pypa/hatch/issues/1452):
```toml
default-args = []
extra-args = ["--doctest-modules", "-m not slow"]
```

leclairm (changes requested):
Fix the doctest issue, potentially with my proposition. Otherwise all good.

.github/workflows/ci.yaml
```diff
@@ -38,7 +38,7 @@ jobs:
       env:
         PYTEST_ADDOPTS: "--durations=0"
       run: |
-        hatch test --cover
+        hatch test -m "" --parallel --cover # override default exclusion of slow tests with -m ""
```
agoscinski, line 41: ah super nice that hatch already comes with pytest-xdist and coverage

agoscinski (approved):
I am happy with it but I did not rigorously review it, since @leclairm already had a look

DropD:
> Finally found workarounds [there](https://github.com/pypa/hatch/issues/1452) and [there](https://github.com/pypa/hatch/issues/1551). This seems to work:
> 
> ```toml
> default-args = []
> extra-args = ["--doctest-modules", "-m not slow"]
> ```

I had deleted that line without checking whether it was needed. Turns out it was, thanks for finding the source of the problem, @leclairm.

leclairm (approved):

PR #118, opened 2025-02-03
infra: Static typing on CI

DropD:


src/sirocco/parsing/_yaml_data_models.py
```diff
@@ -192,53 +192,43 @@ class ConfigCycleTaskOutput(_NamedBaseModel):
...
         if values is None:
             return inputs
         for value in values:
             if isinstance(value, str):
-                inputs.append({value: None})
+                inputs.append(cls(name=value))
             elif isinstance(value, dict):
-                inputs.append(value)
+                inputs.append(cls(**value))
+            else:
```
leclairm, line 210: Ensure idempotency 
```suggestion
            elif not isinstance(value, NAMED_BASE_T):
```

leclairm (changes requested):
Handle the `NAMED_BASE_T` case in `convert_named_model_list`

src/sirocco/parsing/_yaml_data_models.py
DropD: Keep in mind `NAMED_BASE_T` is not a type, but a type var. We will be checking for instances of `_NamedBaseModel` instead.

DropD: I also rewrote the if clause as a match clause.

PR #121, opened 2025-02-06
Refactor dates handling

leclairm:
Fix #119, #97 and incidentally #108 

Short summary: essentially remove as many `None` as possible from date related attributes in order to make the code more understandable.

### Methods
- Insert dedicated classes to test types instead of testing for `None`
- Move some logic to the yaml models although it was already argued against once
- Introduce `cycling` and `target_cycle` keywords in the config format

### Parts to treat
- [x] Previous `_WhenBasemodel`
- [x] ConfigCycle

PR #122, opened 2025-02-11
Improve yaml models: ConfigTasks

### LINKED ISSUE #84: Separate namelist validation in`ConfigIconTask` to dedicated namelist class
Is PR #68 we introduce the field validator for namelist into `ConfigIconTask`. The reason why we cannot put it in separate class with its own validator but have to iterate here is described here
https://github.com/C2SM/Sirocco/blob/3e4da2421282c913d6f3b01d42babdd01ac9cbef/src/sirocco/parsing/_yaml_data_models.py#L393-L395
However this might be fixable but it is for the moment too time consuming what the right way is to do as pydantic is  complex due to its flexibility.

DropD:
Fix #84:

- Introduce a pydantic validated namelists model `ConfigNamelist`
- Refactor `_NamedBaseModel` key extraction class method to a free function for reuse in `ConfigNamelist`
- Simplify `ConfigIconTask` validator
- allow `_tasks.Task` subclasses to hook into how they are created from the equivalent `config.ConfigTask` subclass.
  - required to convert namelist model to data class in `config.ConfigIconTask` -> `_task.IconTask`.

General Improvements:
 
- More doctests
- rename `_yaml_data_models.py` -> `yaml_data_models.py` to reduce conflicts between merges of #121 and #122


src/sirocco/core/_tasks/icon_task.py
```diff
@@ -93,3 +97,17 @@ def section_index(section_name) -> tuple[str, int | None]:
         if m := multi_section_pattern.match(section_name):
             return m.group(1), int(m.group(2)) - 1
         return section_name, None
+
+    @classmethod
+    def build_from_config(cls: type[Self], config: ConfigTask, **kwargs: Any) -> Self:
```
GeigerJ2, line 102: Also, just for my understanding: We're supporting Python >= 3.10, but `Self` type was introduced in 3.11. Is that an issue, or can we use newer Python features in type hinting? According to ChatGPT:

"""
You're correct—Self was introduced in Python 3.11, not 3.12. Since your package claims to support Python 3.10+, using Self directly would be a problem for users on 3.10.

Can you use newer Python concepts in type annotations?
Yes! Starting with PEP 563 (Python 3.7) and PEP 649 (Python 3.10), Python supports postponed evaluation of annotations, meaning type hints are stored as strings instead of being evaluated immediately. This allows you to use newer type hints while keeping compatibility with older versions, as long as they aren't used at runtime.

Solution:
If you add this import at the top of your module:
`from __future__ import annotations`
Then the Self annotation will be treated as a string ("Self") and won't cause syntax errors in Python 3.10. However, tools like mypy will still understand it.
"""

src/sirocco/parsing/yaml_data_models.py
```diff
@@ -37,6 +37,29 @@ def list_not_empty(value: list[ITEM_T]) -> list[ITEM_T]:
     return value
 
 
+def extract_merge_key_as_value(data: Any, new_key: str = "name") -> Any:
+    if not isinstance(data, dict):
+        return data
+    if len(data) == 1:
+        key, value = next(iter(data.items()))
+        match key:
```
GeigerJ2, line 45: Interesting. Haven't really used such `match`/`case` constructs before, but makes the structure clearer than using `isinstance`.

src/sirocco/parsing/yaml_data_models.py
```diff
@@ -298,7 +304,7 @@ class ConfigBaseTaskSpecs:
 
 class ConfigBaseTask(_NamedBaseModel, ConfigBaseTaskSpecs):
     """
-    config for genric task, no plugin specifics
+    Config for genric task, no plugin specifics.
```
GeigerJ2, line 307:
```suggestion
    Config for generic task, no plugin specifics.
```

GeigerJ2 (approved):
LGTM. With the explanations from the meeting, very clear. Just some comments to aid my own understanding. Thanks!

src/sirocco/core/_tasks/icon_task.py
DropD: GPT is correct in that it was introduced later and in that in our case it works (at runtime) because it is treated as a string. Here's how that applies to Sirocco:

For now, we have set the hatch project to run type analysis on 3.12 and we use the future annotations. Therefore it is type checked as if it was supporting 3.12+ (ok at analysis time), while the interpreter treats all the type hints as just strings (ok at run time). 

If we were to lower that to 3.10, we would have to import `Self` from `typing_extensions` (making that a dependency), if we choose to keep using it. It would also affect other things (not touched in this PR) like `class Store[T]`, which would have to turn into `class Store(typing.Generic[T])`. The backports in `typing_extensions` can catch some of this but not syntax changes like `class Store[T]` (which I am honestly suprised that it works at runtime).

src/sirocco/parsing/yaml_data_models.py
DropD: I agree. Now, if there was a `match` expression, the result of which can be assigned to a variable (like in Rust), I would love it even more...

src/sirocco/core/_tasks/icon_task.py
```diff
@@ -3,15 +3,19 @@
...
+from typing import TYPE_CHECKING, Any, Self
 
 import f90nml
 
 from sirocco.core.graph_items import Task
-from sirocco.parsing._yaml_data_models import ConfigIconTaskSpecs
+from sirocco.parsing import _yaml_data_models as models
+
+if TYPE_CHECKING:
+    from sirocco.parsing._yaml_data_models import ConfigTask
```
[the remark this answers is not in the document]
DropD: I might as well clean that up.

PR #126, opened 2025-03-14
Ensure proper workgraph resolution

leclairm:
Make sure workgraph items are created before using them. This PR fixes the fact that, currently, input nodes might not be already created before linking them. At the moment, this depends on the order they've been specified in the config file which should not be required.


agoscinski (commented):
Are you sure about it? We first add all the available on init inputs here https://github.com/C2SM/Sirocco/blob/060e5f984640daf0e1570111db5d03290fb9aaec/src/sirocco/workgraph.py#L83C14-L83C33
 then in 
https://github.com/C2SM/Sirocco/blob/060e5f984640daf0e1570111db5d03290fb9aaec/src/sirocco/workgraph.py#L84
we create output nodes
https://github.com/C2SM/Sirocco/blob/060e5f984640daf0e1570111db5d03290fb9aaec/src/sirocco/workgraph.py#L173
before linking any of the inputs
https://github.com/C2SM/Sirocco/blob/060e5f984640daf0e1570111db5d03290fb9aaec/src/sirocco/workgraph.py#L175
It is a bit hard to read the actual change of the order with all the refactoring of the functions. Can you add a unittest that would help me and is anyway necessary?

agoscinski (changes requested):
notifier

leclairm:
> Are you sure about it?

Yes, see this comment on the other PR: https://github.com/C2SM/Sirocco/pull/125#discussion_r1998267303

I'm sorry, I think this PR hides a very simple fix - breaking the loop over tasks - in a lot of rewriting that just made the overall code feel more readable to me.

src/sirocco/workgraph.py
```diff
@@ -154,6 +155,12 @@ def socket_from_core(self, core_generated_data: core.GeneratedData) -> TaskSocke
     def task_from_core(self, core_task: core.Task) -> aiida_workgraph.Task:
         return self._aiida_task_nodes[self.get_aiida_label_from_graph_item(core_task)]
 
+    def _add_available_data(self):
+        """Adds the available data on initialization to the workgraph"""
+        for data in self._core_workflow.data:
+            if isinstance(data, core.AvailableData):
+                self._add_aiida_input_data_node(data)
+
```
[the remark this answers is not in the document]
leclairm: I have grouped all helper methods to access nodes together on purpose, I don't think it would make sense to move this function back in the middle of the helper ones.

src/sirocco/workgraph.py
```diff
@@ -183,121 +190,115 @@ def _add_tasks(self):
...
+                elif isinstance(input_, core.GeneratedData):
+                    self._workgraph.add_link(
+                        self.socket_from_core(input_), workgraph_task.inputs[f"nodes.{input_label}"]
+                    )
+                else:
+                    raise TypeError
+
+    def _link_wait_on_to_tasks(self):
+        for task in self._core_workflow.tasks:
+            self.task_from_core(task).wait = [self.task_from_core(wt) for wt in task.wait_on]
```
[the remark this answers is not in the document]
GeigerJ2: OK, nvm. With this, from #96, `waiting_on` actually works:

```python
    def _link_wait_on_to_task(self, task: core.Task):
        label = AiidaWorkGraph.get_aiida_label_from_graph_item(task)
        workgraph_task = self._aiida_task_nodes[label]
        workgraph_task.waiting_on.clear()
        for wait_on in task.wait_on:
            wait_on_task_label = AiidaWorkGraph.get_aiida_label_from_graph_item(wait_on)
            workgraph_task.waiting_on.add(self._aiida_task_nodes[wait_on_task_label])

    def _link_wait_on_to_tasks(self):
        for task in self._core_workflow.tasks:
            self._link_wait_on_to_task(task=task)
```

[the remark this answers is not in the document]
leclairm: I saw that as well because pyright complains about `.wait` but I'm not familiar enough with aiida-workgraph so I thought it was on purpose. Now if you see it doesn't show up in the GUI, it might be an actual pb.

src/sirocco/workgraph.py
```diff
@@ -183,132 +190,122 @@ def _add_tasks(self):
...
-        """Links the output to the workgraph task."""
+        for task in self._core_workflow.tasks:
+            if isinstance(task, core.ShellTask):
+                workgraph_task = self.task_from_core(task)
+                if (workgraph_task_arguments := workgraph_task.inputs.arguments) is None:
+                    msg = (
+                        f"Workgraph task {workgraph_task.name!r} did not initialize arguments nodes in the workgraph "
+                        f"before linking. This is a bug in the code, please contact developers."
+                    )
+                    raise ValueError(msg)
```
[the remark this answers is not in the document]
leclairm: Isn't it misleading? I thought sockets in AiiDA were stg else. Here we update the arguments string of the shell task. Is that also considered a socket?

src/sirocco/workgraph.py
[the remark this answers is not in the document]
leclairm: > We can merge this PR here, then rebase and adapt #96, and merge that, as well?

Is #96 still useful if we fix it here?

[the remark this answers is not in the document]
GeigerJ2: For me, either is fine. We can fix it here and close #96. History would then be not as clean, as we mix two changes here, but I think at this early stage of the code base, that's still fine. Otherwise, the approach I mentioned is also fine.

GeigerJ2 (approved):
After discussion with @agoscinski, we agreed to leave the `waiting_on` fix out of this PR, and instead make another PR to fix it after, to keep the history cleaner. From my side, this can be merged now.

src/sirocco/workgraph.py
```diff
@@ -183,121 +190,116 @@ def _add_tasks(self):
...
-    def _link_wait_on_to_task(self, task: core.Task):
-        self.task_from_core(task).wait = [self.task_from_core(wt) for wt in task.wait_on]
+        # NOTE: We don't pass the `nodes` dictionary here, as then we would need to have the sockets available when
+        # we create the task. Instead, they are being updated via the WG internals when linking inputs/outputs to
+        # tasks
+        workgraph_task = self._workgraph.add_task(
+            "ShellJob",
+            name=label,
+            command=command,
+            arguments="__ERROR_NOT_SET__",
```
agoscinski, line 243: But you can have commands without arguments?

src/sirocco/workgraph.py
```diff
@@ -84,6 +84,13 @@ def __init__(self, core_workflow: core.Workflow):
 
         self._add_available_data()
         self._add_tasks()
+        # Creating output sockets requires the corresponding task to already exist
+        self._link_output_nodes_to_tasks()
+        # linking inputs must come after outputs to make sure they were created,
+        # either as data or as socket, before linking
+        self._link_input_nodes_to_tasks()
+        self._set_shelljob_arguments()
```
agoscinski, line 92: `arguments` are part of the the shelljob inputs that is why it was part of the input nodes `_link_inputs_to_ports`. Also I think it is better to have task specific function names after the isinstance check on the task type and not on this higher level

src/sirocco/workgraph.py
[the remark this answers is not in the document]
agoscinski: Hm... I am not really feeling this is moving us to a clearer API but I also don't think it is important to argue about now. Fine for me.

src/sirocco/workgraph.py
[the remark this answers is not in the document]
agoscinski: The argument string is converted after initialization of the task to a Socket. Therefore it is not also accessed like `workgraph_task.inputs.arguments` and not  `workgraph_task.arguments`. As Julian said it is a placeholder to populate when the workgraph is run as not all inputs are available during init. The value can be accessed with `socket.value`.

If it is just about making the line clearer by documenting the type you can also do `workgraph_task_arguments: SocketAny` one line before to keep the name shorter.

agoscinski (changes requested):
Could we disentangle the fix plus tests from the refactor to another PR?

src/sirocco/workgraph.py
leclairm: > arguments are part of the the shelljob inputs

Ok, I had the Sirocco perspective more than the workgraph one. Building the command line sounded a very different action to me than linking the input data nodes. Also, the later has to be done for any task, while the former is only happening for shelljob.

> Also I think it is better to have task specific function after the isinstance check

I agree. I wanted to avoid too much refactoring but I could do that and but all the loops over `self._core_workflow.tasks` and `self._core_workflow.data` at the beginning of `__init__`. That would also make the algorithm clearer.

leclairm: more prcisely, I was thinking about such a sequence in `__init__`:
```python
        # create input data nodes
        for data in self._core_workflow.data:
            if isinstance(data, core.AvailableData):
                self._add_aiida_input_data_node(data)
                
        # create workgraph task nodes and output sockets
        for task in self._core_workflow.tasks:
            # Create the workgraph task itself
            if isinstance(task, core.ShellTask):
                self._create_shell_task_node(task)
            elif isinstance(task, core.IconTask):
                exc = "IconTask not implemented yet."
                raise NotImplementedError(exc)
            else:
                exc = f"Task: {task.name} not implemented yet."
                raise NotImplementedError(exc)
            # Create and link corresponding output sockets
            for output in task.outputs:
                self._link_output_node_to_task(task, output)

        # link input nodes to workgraph tasks
        for task in self._core_workflow.tasks:
            for input_ in task.input_data_nodes():
                self._link_input_node_to_task(task, input_)

        # set shelljob arguments
        for task in self._core_workflow.tasks:
            if isinstance(task, core.ShellTask):
                self._set_shelljob_arguments(task)

        # link wait on to workgraph tasks
        for task in self._core_workflow.tasks:
            self._link_wait_on_to_task(task)
```

src/sirocco/workgraph.py
agoscinski: But if it is just for internal validation that a function is invoked, it is better to create a test that checks if the argument is empty after construction rather than creating this placeholder. When there is a bug this placeholder does not really help the user to debug it while a test actually validates the behavior. This seems something temporary you add for debugging.

src/sirocco/workgraph.py
agoscinski: I am fine with that. Just a note `_link_input_node_to_task` would need to also be separately implemented for icon because `shelljob` uses a dynamic namespace called `nodes` to contain all data node (see e.g. https://github.com/C2SM/Sirocco/blob/79b3be81bf144e41346aa3b009777b8f830fa0c2/src/sirocco/workgraph.py#L271). While in icon ever input is "static" or you could say "hardcoded" (by being defined in aiida-icon) so we directly can set the value in the corresponding socket (`task.inputs.icon_namlist` instead of `task.inputs.nodes.icon_namlist`)

src/sirocco/workgraph.py
agoscinski: I would be okay to mark this line with an issue that we should replace this placeholder with a test for the arguments being not empty after initialization to not block the PR (since it it might be not a super trivial test that needs just 5 minutes to implement)

src/sirocco/workgraph.py
leclairm: All right Thanks for the heads up. Then I'll rewrite this way.

src/sirocco/workgraph.py
```diff
@@ -177,87 +202,84 @@ def _add_aiida_input_data_node(self, data: core.Data):
...
 
-        This includes the linking of all input and output nodes, the arguments and wait_on tasks
-        """
-        for task in self._core_workflow.tasks:
-            self._create_task_node(task)
+        if isinstance(task, core.IconTask):
+            msg = "method not implemented yet for Icon tasks"
+        else:
+            msg = f"method not implemented for task type {type(task)}"
+        raise NotImplementedError(msg)
```
GeigerJ2, line 213: `singledispatchmethod` shouldn't require the `isinstance` check, as that's exactly what it is supposed to replace. Typically, this case distinction would be made via the registered methods. I propose to change to (cannot make it as a suggestion for whichever reason):
```python
@functools.singledispatchmethod
def create_task_node(self, task):
    """dispatch creating task nodes based on task type"""
    msg = f"method not implemented for task type {type(task)}"
    raise NotImplementedError(msg)

@create_task_node.register
def _create_icon_task_node(self, task: core.IconTask):
    msg = "method not implemented yet for Icon tasks"
    raise NotImplementedError(msg)

@create_task_node.register
def _create_shell_task_node(self, task: core.ShellTask):
	...
```

src/sirocco/workgraph.py
```diff
@@ -177,87 +202,84 @@ def _add_aiida_input_data_node(self, data: core.Data):
...
+    @functools.singledispatchmethod
+    def _link_input_node_to_task(self, task: core.Task, input_: core.Data):  # noqa: ARG002
+        """ "Dispatch linking input to task based on task type"""
+
+        msg = f"method not implemented for task type {type(task)}"
+        raise NotImplementedError(msg)
+
+    @_link_input_node_to_task.register
+    def _link_input_node_to_shelltask(self, task: core.ShellTask, input_: core.Data):
+        """Links the input to the workgraph shell task."""
```
GeigerJ2, line 281: Not sure if we need the `singledispatch` here, as we have only one registered method, so far. But fine to keep it in anticipation of `IconTask`!

GeigerJ2 (approved):
Good to go from my side. Thanks, @leclairm! Just added a minor comment regarding `singledispatch`.

src/sirocco/workgraph.py
leclairm: Yes it was the aim.

src/sirocco/workgraph.py
leclairm: That works as well. It was just a quick way to raise a slightly different error message for the icon task. And I didn't put the placeholder in the `link_input_xxx` case so I didn't do it here either. This will anyway go as soon as the icon task is there.

PR #136, opened 2025-03-27
WIP: Remote submission

GeigerJ2:


src/sirocco/workgraph.py
```diff
@@ -317,6 +322,26 @@ def _set_shelljob_arguments(self, task: core.ShellTask):
...
         workgraph_task_arguments.value = arguments
 
+    def _set_shelljob_filenames(self, task: core.ShellTask):
+        """set AiiDA ShellJob filenames for AvailableData entities"""
+        workgraph_task = self.task_from_core(task)
+
+        filenames = {}
+        input_sockets = workgraph_task.inputs.nodes
+
+        for input_socket in input_sockets:
```
GeigerJ2, line 332: Check that this doesn't mess with `SinglefileData`.

GeigerJ2, line 332: Check that this doesn't mess with `SinglefileData`.

GeigerJ2, line 332: Check that this doesn't mess with `SinglefileData`.

src/sirocco/workgraph.py
```diff
@@ -317,6 +322,26 @@ def _set_shelljob_arguments(self, task: core.ShellTask):
         _, arguments = self.split_cmd_arg(task.resolve_ports(input_labels))
         workgraph_task_arguments.value = arguments
 
+    def _set_shelljob_filenames(self, task: core.ShellTask):
+        """set AiiDA ShellJob filenames for AvailableData entities"""
+        workgraph_task = self.task_from_core(task)
+
+        filenames = {}
+        input_sockets = workgraph_task.inputs.nodes
```
leclairm, line 330: The current implementation is a bit hacky and there already are helper functions for accessing the aiida nodes from the core nodes.
```suggestion
        for input_ in task.inputs:
            if isinstance(input_, core.AvailableData):
                input_socket = self.data_from_core(input_)
                filenames[input_socket._name] = Path(input_.src).name
```

src/sirocco/workgraph.py
```diff
@@ -317,6 +322,26 @@ def _set_shelljob_arguments(self, task: core.ShellTask):
...
+
+        for input_socket in input_sockets:
+            # No public `name` attribute available for SocketAny
+            input_socket_name = input_socket._name  # noqa: SLF001
+            if input_socket_name in self._aiida_data_nodes:
+                # Pass empty tuple as second argument to __getitem__ of the Store
+                # as AvailableData don't have parameters (or do they??)
+                core_data = self._core_workflow.data[input_socket_name, {}]
+                if isinstance(core_data, core.AvailableData):
+                    filenames[input_socket_name] = Path(core_data.src).name
```
leclairm, line 340:
```suggestion
        for input_ in task.inputs.values():
            if isinstance(input_, core.AvailableData):
                input_socket = self.data_from_core(input_)
                filenames[input_socket._name] = Path(input_.src).name
```

GeigerJ2: Hi @leclairm! Thanks for the suggestion. I had applied it, but the test `test_run_workgraph_small_remote` started failing, and during debugging I saw that with the proposed code, `filenames` is always empty. I'm currently still checking what the problematic changes are, but I might revert the commit.

GeigerJ2: OK, I fixed the function. Reads simpler and less hacky now. Thanks!

pyproject.toml
```diff
@@ -35,7 +35,8 @@ dependencies = [
   "termcolor",
   "pygraphviz",
   "lxml",
-  "f90nml"
+  "f90nml",
+  "aiida-shell @ git+https://github.com/sphuber/aiida-shell.git@fix/105/handle-remote-data-argument-placeholders",
```
GeigerJ2, line 39: Need to pin here before the PR is merged and released.

src/sirocco/workgraph.py
```diff
@@ -56,12 +56,22 @@ def _prepare_for_shell_task(task: dict, inputs: dict) -> dict:
...
+        "stdout",
+        "stderr",
+    }
     task_outputs = set(task["outputs"].keys())
     task_outputs = task_outputs.union(set(inputs.pop("outputs", [])))
     missing_outputs = task_outputs.difference(default_outputs)
     inputs["outputs"] = list(missing_outputs)
     # Workaround ends here
+    # NOTE: Hardcoded for now, possibly make user-facing option
+    inputs["metadata"]["options"] = {"use_symlinks": True}
```
agoscinski, line 74: Do we need to do it in the workgraph workaround? We also could also define the metadata in the task creation see `_create_shell_task_node`

src/sirocco/workgraph.py
```diff
@@ -317,6 +322,23 @@ def _set_shelljob_arguments(self, task: core.ShellTask):
         _, arguments = self.split_cmd_arg(task.resolve_ports(input_labels))
         workgraph_task_arguments.value = arguments
 
+    def _set_shelljob_filenames(self, task: core.ShellTask):
+        """set AiiDA ShellJob filenames for AvailableData entities"""
+
+        filenames = {}
+
+        for input_list in task.inputs.values():
```
agoscinski, line 330: inputs can be a list of inputs because for shell task we can assign multiple inputs for one port and the list returns you all the inputs for this one port. I don't think we stress test this much so most of the time the input_list has just one element.
```suggestion
        for input_list in task.input_data_nodes:
```

tests/test_wc_workflow.py
```diff
@@ -29,18 +31,66 @@ def test_vizgraph(config_paths):
     "config_case",
     [
         "small",
+    ],
+)
+def test_run_workgraph_small_remote(config_case, config_paths, aiida_localhost, tmp_path):  # noqa: ARG001  # config_case is overridden
```
agoscinski, line 36: Can't you just change one of the `config.yml` by adding a computure to one of the data nodes there? I am actually not sure how this test tests `RemoteData` since I don't see anywhere a computer information. If you really want to test if something is copied maybe you do it by passing a symlink and in the script you test if you got a symlink or if the symlink got correctly resolved.

tests/cases/small/config/config.yml
```diff
@@ -35,9 +35,12 @@ tasks:
   - icon:
       computer: localhost
       plugin: shell
+      # If `src` not absolute, this ends up to be relative to the root directory of the config file -> Should also be
+      # solved by registering `Code`s in AiiDA for the required scripts
+      # See: https://github.com/C2SM/Sirocco/issues/127
+      # And: https://github.com/C2SM/Sirocco/issues/60
```
agoscinski, line 41: So pydantic has the member variable `description` for their `Field`, AFAIK this is exported into the schema. So better you add this there rather than in a test file https://github.com/C2SM/Sirocco/blob/807a6fb3cd045c38dc7406906903a76542aaced5/src/sirocco/parsing/yaml_data_models.py#L282

agoscinski (changes requested):

pyproject.toml
agoscinski: Do we need this change in aiida-shell or would this just make our life easier?

src/sirocco/workgraph.py
agoscinski: I think you need to handle all inputs separate as these can be potentially different files. Note that currently our code is also buggy in this regard. If we pass the outputs of a parametrized shell task to another task aiida will copy the file to the working directory. Since the output filename is the same for each parameter they will be overwritten. So only one file exists in the working directory of the next tasks that gets this as input. The solution is to give the outputs unique names.

src/sirocco/workgraph.py
```diff
@@ -317,6 +322,23 @@ def _set_shelljob_arguments(self, task: core.ShellTask):
...
+            # Some empty lists appear here?
+            if not input_list:
+                continue
```
agoscinski, line 333: I think this happens when you use the `when` keyword. So for example, for the icon restart file that is not present in the first cycle iteration. This is represented as empty list in the inputs `'restart': []` for that datetime. I think since we need to iterate through all inputs (as suggested above), this will be not needed anymore.

src/sirocco/parsing/yaml_data_models.py
```diff
@@ -281,7 +281,6 @@ class ConfigShellTaskSpecs:
     sep_pattern: ClassVar[re.Pattern] = field(default=re.compile(r"\[sep=(.+)\]"), repr=False)
     src: str | None = None
     command: str
-    env_source_files: list[str] = field(default_factory=list)
```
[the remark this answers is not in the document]
GeigerJ2: I discussed this with @leclairm. For the previous PR title "WIP: Remote Submission", it was fitting, as it was more general, and the environment source files would break remote submission anyway. Now, as I better traced back the other root problem with submitting remotely that this PR mainly addresses now, and adapted the title, it's not fitting 100% anymore. While we still decided to keep the change here, I can also move it to another PR.

pyproject.toml
GeigerJ2: Yeah, this is required. Without it, the instructions will always be appended by `/*`, so will always fail for our `RemoteData` files, see here:
https://github.com/sphuber/aiida-shell/blob/189df631759eb07e574bfb5a5be2843ea532ec9d/src/aiida_shell/calculations/shell.py#L341
We could still hack our way around it by using the parent of the `src` path, but then, still, all files in the directory will be linked. I think it's better if we keep it pinned for now, and expect the PR to aiida-shell to be merged and released.

src/sirocco/workgraph.py
GeigerJ2: Fair, good point! Will adapt.

tests/cases/small/config/config.yml
GeigerJ2: Thanks! moved it there.

src/sirocco/parsing/yaml_data_models.py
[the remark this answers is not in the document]
GeigerJ2: OK, I removed it from this PR to keep the history cleaner

tests/test_wc_workflow.py
GeigerJ2: I reverted the changes to the YAML files now, and instead added more isolated tests

agoscinski:
Can you rebase so we can run the tests?

GeigerJ2:
Hi @agoscinski, I merged from main and resolved the conflicts. Been recently still trying to fix the renaming/filenames also for parametrized task, see [here](https://github.com/GeigerJ2/Sirocco/blob/7c9ed55afe393e1f49645660338f71c7d1262011/tests/test_workgraph.py#L271), but I don't think this works fully yet...

pyproject.toml
agoscinski: Can't we monkeypatch `ShellJob` and pin the exact version? When we start to include git hashes in our dependencies I am afraid we run into problems later on. Dependency resolvement becomes a bit intransparent and unpredictable, e.g. we pin a hash while wrokgraph pins a version.

sphuber: I finally released this fix in v0.8.1 just now. Had forgotten about it, sorry for the delay

GeigerJ2: Cheers, thanks a lot @sphuber! Just in time that this PR actually works and can be reviewed 🚀

pyproject.toml
```diff
@@ -46,7 +47,7 @@ Changelog = "https://github.com/C2SM/Sirocco/blob/main/CHANGELOG.md"
 
 [tool.pytest.ini_options]
 # Configuration for [pytest](https://docs.pytest.org)
-addopts = "--pdbcls=IPython.terminal.debugger:TerminalPdb"
+addopts = "-s --pdbcls=IPython.terminal.debugger:TerminalPdb"
```
GeigerJ2, line 50: `-s` disables output capturing, and allows for setting breakpoints in test code

PR #139, opened 2025-04-24
Add termcolor types to deps

agoscinski:


agoscinski:
Yes we can, just tried to respect what the person did by introducing this import of Color. This way you still document that this is a Color type (even though a color type is just str). Whatever you want, I just want to merge this.

src/sirocco/core/graph_items.py
```diff
@@ -2,7 +2,7 @@
 
 from dataclasses import dataclass, field
 from itertools import chain, product
-from typing import TYPE_CHECKING, Any, ClassVar, Self, TypeVar, cast
+from typing import TYPE_CHECKING, Any, ClassVar, Self, TypeAlias, TypeVar, cast
```
agoscinski, line 5:
```suggestion
from typing import TYPE_CHECKING, Any, ClassVar, Self, TypeVar, cast
```

src/sirocco/core/graph_items.py
```diff
@@ -27,6 +25,9 @@
     )
 
 
+Color: TypeAlias = str
+
+
```
agoscinski, line 30:
```suggestion
```

agoscinski, line 30:
```suggestion
```

PR #141, opened 2025-04-25
Include ports to `ConfigCycleTaskOutput` and `Data`

agoscinski:
Outputs now also can have an optional port. This change is required for icon task that uses the `IconCalculation` aiida calcjob to run icon. Unlike the aiida `ShellJob` that creates output ports dynamically from the `output.src` information, the `IconCalculation` defines fixed output ports for each type of output. For example to reference the outputted restart file we need to write a config of the form.

```yaml
  outputs:
    - restart:
        port: restart_file
```

The disadvantage of this design to support ports for outputs is that one can also specify ports for the `ShellTask` outputs where it is not needed. My solution to this problem is to notify the user with a warning that this does not have any effect.

We need the `graph_spec.outputs` information of the spec in the children classes of the Task (`ShellTask` and `IconTask` in a future PR) to resolve the data port differently. To have this information available in the children class instead of passing the `graph_spec.outpus` where the port information resides to the __init__ constructor, I  added the port information to the `Data` object that is already passed to the children class. This is an easier approach, since we use dataclass decorator for the children classes creating a custom __init__ constructor which is a bit cumbersome to overload. To be more consistent I also adapted the inputs to the same approach so we use the same type for inputs and outputs. I added a method to obtain inputs by ports that can be made more efficient by caching and a O(n log n) algorithm, if this this becomes ever a bottleneck

### DIFF
--- a/src/sirocco/core/graph_items.py
+++ b/src/sirocco/core/graph_items.py
@@ -71,7 +71,7 @@ class Task(ConfigBaseTaskSpecs, GraphItem):
     color: ClassVar[str] = field(default="light_red", repr=False)
 
     inputs: dict[str, list[Data]] = field(default_factory=dict)
-    outputs: list[Data] = field(default_factory=list)
+    outputs: dict[str | None, list[Data]] = field(default_factory=dict)
     wait_on: list[Task] = field(default_factory=list)
     config_rootdir: Path
     cycle_point: CyclePoint
@@ -88,6 +88,9 @@ class Task(ConfigBaseTaskSpecs, GraphItem):
     def input_data_nodes(self) -> Iterator[Data]:
         yield from chain(*self.inputs.values())
 
+    def output_data_nodes(self) -> Iterator[Data]:
+        yield from chain(*self.outputs.values())
+
     @classmethod
     def from_config(
         cls: type[Self],
@@ -103,7 +106,13 @@ class Task(ConfigBaseTaskSpecs, GraphItem):
             if input_spec.port not in inputs:
                 inputs[input_spec.port] = []
             inputs[input_spec.port].extend(datastore.iter_from_cycle_spec(input_spec, coordinates))
-        outputs = [datastore[output_spec.name, coordinates] for output_spec in graph_spec.outputs]
+
+        outputs: dict[str | None, list[Data]] = {}
+        for output_spec in graph_spec.outputs:
+            if output_spec.port not in outputs:
+                outputs[output_spec.port] = []
+            outputs[output_spec.port].append(datastore[output_spec.name, coordinates])
+
         if (plugin_cls := Task.plugin_classes.get(type(config).plugin, None)) is None:
             msg = f"Plugin {type(config).plugin!r} is not supported."
             raise ValueError(msg)
--- a/src/sirocco/parsing/yaml_data_models.py
+++ b/src/sirocco/parsing/yaml_data_models.py
@@ -173,6 +173,8 @@ class ConfigCycleTaskOutput(_NamedBaseModel):
     To create an instance of an output in a task in a cycle defined in a workflow file.
     """
 
+    port: str | None = None
+
 
 NAMED_BASE_T = typing.TypeVar("NAMED_BASE_T", bound=_NamedBaseModel)
 
--- a/src/sirocco/pretty_print.py
+++ b/src/sirocco/pretty_print.py
@@ -139,7 +139,7 @@ class PrettyPrinter:
             sections.append(
                 self.as_block(
                     "output",
-                    "\n".join(self.as_item(self.format_basic(output)) for output in obj.outputs),
+                    "\n".join(self.as_item(self.format_basic(output)) for output in obj.output_data_nodes()),
                 )
             )
         if obj.wait_on:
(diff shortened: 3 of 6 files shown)


agoscinski (commented):
remove my TODOs for other PRs

leclairm (changes requested):
While I understand the need to attach a port information to an output data node, I think this implementation fundamentally breaks what a port actually is. It's giving a role to an object, it's not a property of that object but of the task. What makes the full information is the combination of the port and the object and the target port may vary depending on which task the object is linked to, now as input or output. The only case where this implementation could work is if a data node is always attached to the same port for any task, be it  as input or output. This works for restarts if the port name is the same in the input and output cases but is much too restrictive.

I suggest instead to take the inverse approach and mimic what was done for inputs: A task has an `output` dictionary mapping ports to data nodes. In the case of `ShellTask`, we could map all the output data nodes to a `"default"` port.

agoscinski:
As I wrote in my first post (see last paragraph), I followed first the approach you suggested but this approach will result in more complicated code (-> less readable code). You will need a custom `__init__` constructor because `@dataclass` creates an init constructor which we cannot anymore use as we need to also pass the spec there. AFAIK `@dataclass` does the magic to pass the right input arguments to right parent class. Thanks for pointing out the concern that one data node can be associated multiple tasks for different ports, but the change is within the Task associated to a cycle so I don't understand how this can be a concern. We are already handling at this point data nodes for specific coordinates (data + parameters). If different tasks can share the same data objects, I find this a bit concerning since the ownership of the data is not clear which can result in hard-to-debug bugs when one task changes the data and by that effects something in another task.

leclairm:
How do you handle this?
```yaml
cycles:
  - icon_bimonthly:
      cycling:
        start_date: *root_start_date
        stop_date: *root_stop_date
        period: 'P2M'
      tasks:
        - icon:
            inputs:
              - icon_restart:
                  when:
                    after: *root_start_date
                  target_cycle:
                    lag: '-P2M'
                  port: restart
              - [...]
            outputs: [...]
        - store_and_clean:
            inputs:
              - icon_restart:
                  port: my_input
              - [...]
            outputs: [...]
[...]
tasks:
  - store_and_clean:
      plugin: shell
      src: scripts/store_clean.sh
      command: "cleanup.sh --my_opt {PORT::my_input}"
      [..]
```

agoscinski:
I agree with what you wrote, my worry is not about that conceptually the data node is reused for multiple tasks, but rather that the underlying python object is reused for multiple tasks. Even though we reuse the same data node, we should not reuse the same python object as this avoids encapsulation, i.e. I need to always consider that manipulating an object in one task changes an object in other tasks. Further this not even indicated/enforced in the code (e.g. using weakref, immutable dict) so this can easily cause catastrophic bugs if one forgets about it. While we could use this pattern in the code more properly by using immutable dict for example, my question would be why should we use the more complicated pattern if copying is much simpler. You need to maintain much more code to have an implementation that is closer to what conceptually is happening.

```python
from sirocco.parsing.yaml_data_models import ConfigWorkflow, validate_yaml_content
from sirocco.core import Workflow
import textwrap
content = textwrap.dedent(
    '''
    name: minimal
    rootdir: /location/of/config/file
    cycles:
      - minimal_cycle:
         tasks:
           - task_a:
               inputs:
                 - foo:
                     port: arg
               outputs:
                 - bar
           - task_b:
               inputs:
                 - bar:
                     port: arg 
               outputs:
                 - bar
    tasks:
      - task_a:
          plugin: shell
          command: "some_command {PORT::arg}"
      - task_b:
          plugin: shell
          command: "some_command {PORT::arg}"
    data:
      available:
        - foo:
            type: file
            src: foo.txt
      generated:
        - bar:
            type: dir
            src: bar
    '''
)

wf = validate_yaml_content(ConfigWorkflow, content)
cwf = Workflow.from_config_workflow(wf)
tasks = [t for t in cwf.tasks]
# same object
print(id(tasks[0].outputs[0]))
print(id(tasks[1].inputs['arg'][0]))
# Dangerous because one tasks can change things in other tasks. You have kind of a global state because one can only think of tasks as a whole when manipulationg the data objects, 
def foo_local_change(task):
    for output in task.outputs:
        output.name += '-append'
foo_local_change(tasks[0])
# Ooops I also changed data somewhere else
print(tasks[1].inputs['arg'][0].name) # nothing in the code notified me about this shared object 
# --> Harder to keep track of code as you have a shared state through all tasks.
# Do we need to allow this kind of manipulations within a task? I don't think so.

# In principle you could also only store copies
import copy
tasks[0].outputs[0] = copy.deepcopy(tasks[0].outputs[0])
foo_local_change(tasks[0])
print(tasks[1].inputs['arg'][0].name) # no further change

leclairm:
I kind of see that concern but I think such a bug would come from a misunderstanding of what the core data nodes are. They are elements of the internal representation of the workflow graph. Making as many copies of these nodes as they have linked tasks would break the graph structure by losing the connections between the nodes, be it in the workflow IR or in the `AiidaWorkGraph`.

Actually, opposite to your concern, graph nodes might have to be modified and the new piece of information being available to connected nodes. For instance, in the case of a  generated output data node, how would a subsequent task using it as input know the path to if it uses an unrelated copy of that node.

agoscinski:
> Actually, opposite to your concern, graph nodes might have to be modified and the new piece of information being available to connected nodes. 

If we need to manipulated the datastore later on, this would be an argument for the current design. But I am missing the use case for later on manipulating the data nodes (so the data store object), as we only create the workflow metadata and then run it. Maybe you can elaborate a bit what use case you have in mind where the data store object needs modification later on.

leclairm (approved):
I'm just unsure about the optional nature of output ports but we can merge.
Thanks!

PR #142, opened 2025-05-05
Icon namelist refactor

agoscinski:
Refactor of `Namelist` and ¨IconTask`:
- Reduce `IconTask` API to `update_namelists_from_workflow` as other
  functions could be moved to other places or were note needed .
- Separate dumping of namelists from the update process as one can pass
  the namelist by a memory buffer (io.StringIO) in AiiDA.
- Be more consistent with the `Config<CLASS_NAME>Spec` naming convention
  and usage. This class is used to for shared member variables between
  the config file and core representation.

Convert all path variables in config schema to `pathlib.Path`.

### DIFF
--- a/src/sirocco/core/_tasks/icon_task.py
+++ b/src/sirocco/core/_tasks/icon_task.py
@@ -15,45 +15,66 @@ if TYPE_CHECKING:
 
 
 @dataclass(kw_only=True)
-class Namelist(models.ConfigNamelistSpec):
+class IconNamelist(models.ConfigIconNamelistSpec):
+    """A namelist with a name and a path."""
+
     name: str = field(init=False)
 
     def __post_init__(self):
-        self.update_from_path(self.path)
+        self.path = self._validate_namelist_path(self.path)
         self.name = self.path.name
+        self._namelist = f90nml.read(self.path)
+
+    @classmethod
+    def from_config(cls, config: models.ConfigIconNamelist, config_rootdir: Path):
+        path = cls._validate_namelist_path(config.path, config_rootdir)
+        self = cls(path=path)
+        self.update_from_specs(config.specs)
+        return self
 
     @property
-    def content(self) -> f90nml.Namelist:
-        return self._content
-
-    def update_from_path(self, path: Path) -> None:
-        """Updates the internal content from a namelist in path."""
-        self._content = f90nml.read(path)
-        self.path = path
+    def namelist(self) -> f90nml.Namelist:
+        return self._namelist
 
     def update_from_specs(self, specs: dict[str, Any]) -> None:
-        """Updates the internal content from the specs."""
+        """Updates the internal namelist from the specs."""
         for section, params in specs.items():
             section_name, k = self.section_index(section)
             # Create section if non-existent
-            if section_name not in self.content:
+            if section_name not in self.namelist:
                 # NOTE: f90nml will automatially create the corresponding nested f90nml.Namelist
                 #       objects, no need to explicitly use the f90nml.Namelist class constructor
-                self.content[section_name] = {} if k is None else [{}]
+                self.namelist[section_name] = {} if k is None else [{}]
             # Update namelist with user input
             # NOTE: unlike FORTRAN convention, user index starts at 0 as in Python
-            if k == len(self.content[section_name]) + 1:
+            if k == len(self.namelist[section_name]) + 1:
                 # Create additional section if required
-                self.content[section_name][k] = f90nml.Namelist()
-            nml_section = self.content[section_name] if k is None else self.content[section_name][k]
+                self.namelist[section_name][k] = f90nml.Namelist()
+            nml_section = self.namelist[section_name] if k is None else self.namelist[section_name][k]
             nml_section.update(params)
 
-    def dump_content(self, path: Path):
-        import io
+    def dump(self, path: Path):
+        if path.is_file():
+            path.unlink()
+        if path.is_dir():
+            msg = f"Cannot write namelist {path.name} to path {path.name} already exists."
+            raise OSError(msg)
+        self.namelist.write(path)
 
-        with io.StringIO() as buffer:
-            self.content.write(buffer)
-            path.write_text(buffer.getvalue())
+    @staticmethod
+    def _validate_namelist_path(config_namelist_path: Path, config_rootdir: Path | None = None) -> Path:
+        if config_rootdir is None and not config_namelist_path.is_absolute():
+            msg = f"Cannot specify relative path {config_namelist_path} for namelist while the rootdir is None"
+            raise ValueError(msg)
+
+        namelist_path = config_namelist_path if config_rootdir is None else (config_rootdir / config_namelist_path)
+        if not namelist_path.exists():
+            msg = f"Namelist in path {namelist_path} does not exist."
+            raise FileNotFoundError(msg)
+        if not namelist_path.is_file():
+            msg = f"Namelist in path {namelist_path} is not a file."
+            raise OSError(msg)
+        return namelist_path
 
     @staticmethod
     def section_index(section_name) -> tuple[str, int | None]:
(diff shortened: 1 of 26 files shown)


src/sirocco/core/_tasks/icon_task.py
```diff
@@ -95,6 +74,86 @@ def section_index(section_name) -> tuple[str, int | None]:
...
+
+    def update_namelists_from_workflow(self):
+        if not isinstance(self.cycle_point, DateCyclePoint):
+            msg = "ICON task must have a DateCyclePoint"
+            raise TypeError(msg)
+        self.master_namelist.update_from_specs(
+            {
+                "master_time_control_nml": {
+                    "experimentStartDate": self.cycle_point.start_date.isoformat() + "Z",
+                    "experimentStopDate": self.cycle_point.stop_date.isoformat() + "Z",
```
agoscinski, line 138: Shouldn't it rather change the `checkpointtimeintval` and `restarttimeintval`?

agoscinski, line 138: Shouldn't it rather change the `checkpointtimeintval` and `restarttimeintval`?  (Haven't changed this)

tests/cases/parameters/data/config.txt
```diff
@@ -12,7 +12,7 @@ cycles:
             coordinates: {'foo': 0, 'bar': 3.0, 'date': datetime.datetime(2026, 1, 1, 0, 0)}
             cycle point: [2026-01-01 00:00:00 -- 2026-07-01 00:00:00]
             plugin: 'shell'
-            src: 'scripts/icon.py'
+            src: scripts/icon.py
```
agoscinski, line 15: This is due to the change of `src` from str to Path. I can also move this to another PR if requested

src/sirocco/core/_tasks/icon_task.py
```diff
@@ -104,9 +162,28 @@ def build_from_config(cls: type[Self], config: models.ConfigTask, **kwargs: Any)
...
+                msg = f"Namelist in path {namelist_path} is not a file."
+                raise OSError(msg)
+            return namelist_path
+
+        config_rootdir = kwargs["config_rootdir"]
+        namelists = []
+        for config_namelist in config_kwargs["namelists"]:
+            namelist_path = validate_config_namelist_path(config_namelist.path, config_rootdir)
+            namelist = Namelist(path=namelist_path)
+            namelist.update_from_specs(config_namelist.specs)
```
agoscinski, line 183: merge this function to classmethod constructor

src/sirocco/core/_tasks/icon_task.py
```diff
@@ -2,80 +2,58 @@
...
-            if folder is None:
-                folder = (self.config_rootdir / cfg_nml.path).parent
-            suffix = ("_".join([str(p) for p in self.coordinates.values()])).replace(" ", "_")
-            self.core_namelists[name].write(folder / (name + "_" + suffix), force=True)
 
-    def create_workflow_namelists(self, folder=None):
-        self.init_core_namelists()
-        self.update_core_namelists_from_config()
-        self.update_core_namelists_from_workflow()
-        self.dump_core_namelists(folder=folder)
```
agoscinski, line 78: put back into old names, only `init_core_namelists` has been removed and put into the initialization

tests/conftest.py
```diff
@@ -101,7 +103,7 @@ def serialize_nml(config_paths: dict[str, pathlib.Path], workflow: workflow.Work
     nml_refdir = config_paths["txt"].parent / "ICON_namelists"
     for task in workflow.tasks:
         if isinstance(task, core_tasks.icon_task.IconTask):
-            task.create_workflow_namelists(folder=nml_refdir)
+            task.dump_namelists(path=nml_refdir)
```
agoscinski, line 106: include `update_namelists_from_workflow` even here or in the dumping

src/sirocco/core/_tasks/icon_task.py
```diff
@@ -2,80 +2,58 @@
...
+                self.content[section_name][k] = f90nml.Namelist()
+            nml_section = self.content[section_name] if k is None else self.content[section_name][k]
+            nml_section.update(params)
+
+    def dump_content(self, path: Path):
+        import io
+
+        with io.StringIO() as buffer:
+            self.content.write(buffer)
+            path.write_text(buffer.getvalue())
```
agoscinski, line 56: simplify this back to how it was done before as it is harder to read

src/sirocco/core/_tasks/icon_task.py
```diff
@@ -95,6 +73,86 @@ def section_index(section_name) -> tuple[str, int | None]:
...
+                master_namelist = namelist
+                break
+        if master_namelist is None:
+            msg = f"Failed to read master namelists. Could not find {self._MASTER_NAMELIST_NAME!r} in namelists {self.namelists}"
+            raise ValueError(msg)
+        self._master_namelist = master_namelist
+
+        # retrieve model namelist name from master namelist
+        if (master_model_nml := self._master_namelist.content.get("master_model_nml", None)) is None:
+            msg = "No model filename specified in master namelist: Could not find section '&master_model_nml'"
```
agoscinski, line 99: `&master_model_nml` could be also ClassVar[str]

src/sirocco/core/_tasks/icon_task.py
```diff
@@ -2,80 +2,58 @@
...
-            suffix = ("_".join([str(p) for p in self.coordinates.values()])).replace(" ", "_")
-            self.core_namelists[name].write(folder / (name + "_" + suffix), force=True)
 
-    def create_workflow_namelists(self, folder=None):
-        self.init_core_namelists()
-        self.update_core_namelists_from_config()
-        self.update_core_namelists_from_workflow()
-        self.dump_core_namelists(folder=folder)
+@dataclass(kw_only=True)
+class Namelist(models.ConfigNamelistSpec):
```
agoscinski, line 18: Try to subclass f90nml.Namelist otherwise give better name this class e.g. `CoreNamelist`

agoscinski: I tried to inherit directly from f90nml.Namelist but got weird bugs when I accessed the object in `update_from_specs`. Also it makes to me more sense that the namelist itself does not have a name but an icon namelist has always a name (an identfier to be specified). The name is also important for the dumping. So I continued to use the composition over inheritance pattern and changed the name to `IconNamelist`.

src/sirocco/core/_tasks/icon_task.py
agoscinski: Because of renaming it to `IconNamelist` I also us the function name `update_icon_namelists_from_workflow` and `dump_icon_namelists`.  The function `update_core_namelists_from_config` I kept `update_from_specs` because it updates the namelist from a generic dict and is also used for generic dicts during the runtime of the workflow.

src/sirocco/core/_tasks/icon_task.py
agoscinski: done

src/sirocco/core/_tasks/icon_task.py
agoscinski: @leclairm still open question actually if we need to update also the `checkpointtimeintval` and `restarttimeintval`

src/sirocco/core/_tasks/icon_task.py
agoscinski: done

src/sirocco/core/_tasks/icon_task.py
leclairm: Ah yes, true `restarttimeinterval` must be updated with the period. It uses the same ISO standard so we can just use the period directly.

leclairm (changes requested):

src/sirocco/core/_tasks/icon_task.py
leclairm: That's fine. I was fearing we'd have these kind of problems since `f90nml.Namelist` is recursive so modifying its `__init__` method is not trivial.

leclairm:
```python
>>> import f90nml
>>> nml = f90nml.Namelist()
>>> nml['section_1'] = {'a': 1, 'b': 2}
>>> type(nml)
"f90nml.namelist.Namelist"

>>> section = nml['section_1']
>>> type(section)
"f90nml.namelist.Namelist"

```

leclairm:
For some reason my comment on the renaming got lost. I think we should not use `IconNamelist` as the class doesn't contain any ICON specifics, which is good so we can reuse it in another context. Another task plugin for instance. I'd use `NamelistFile` because, actually, the word "namelist" is misused in the community. Technically, it refers to a section in an ASCII file delimited by 2 lines like this
```
&XYZ
name_1 = value_1
name_2 = Value_2
[...]
/
```
and the file can (usually does) contain several of these. So `NamelistFile` better relates to the file containing the namelists.

leclairm (approved):

PR #143, opened 2025-05-06
Add support of icon task in workgraph module

agoscinski:
Added a new test case `small-icon` and renamed test case `small` to `small-shell`. The `small-icon` test case is a copy of `small` replacing the usage of the dummy icon scripts with real icon. Because the `small` use case is still a good test case for the `ShellTask` I kept it.

- Add dispatching for `IconTask` in multiple functions in responsible for creating the WorkGraph.
- Pass `computer` argument to `core.Task` since it is required for the creation of a `IconCaclculation`. In the `small-shell` config the usage of `computer` has been removed temporarly until PR #136 fixes the usage.
- Changes in `core.AvailableData`:
  - Use `config_rootdir` to resolve location of data if relative.
  - The `src` member is now compulsory and validated. This change required to implement the `from_config` constructor for `AvailableData` and `GeneratedData` as the validation does not happen for `GeneratedData`.
- Fixing how `is_restart` is determined: It was not considering the data structure correctly data items. It was only checking the existance of the restart key in the input items, it was, however, not considering that the data structure still lists the input item with an empty list when the `when` keyword is used. Now it validates correctly to `False`.
- In tests we use now `pytest.fixture.usefixtures` when a fixture is not directly used but is required to be executed

For certain changes I though a justification is required I wrote a comment in the comment starting with the string `PRCOMMENT`.

Future PRs:
- Support MPI through config. Now just disabled it because I don't understand where aiida defines `tot_num_mpiprocs`. Also we might want this to become a parameter of the sirocco config since it would be annyoing to create a new computer for this all the time. 
- Fix warning:
```
Warning: [submission of calculation 34] Unable to copy remote resource from ICON/ECHAM6_CldOptProps.nc to ECHAM6_CldOptProps.nc! NOT Stopping but just ignoring!.
Warning: [submission of calculation 34] Unable to copy remote resource from ICON/dmin_wetgrowth_lookup.nc to dmin_wetgrowth_lookup.nc! NOT Stopping but just ignoring!.
```
- An aiida Installed code is created for each icon task but we define the src globally so move this somewhere in the workgraph creation to be only once created per workflow
- We need to separate the children class logic from the parent class. I mean for example that the Task
- It would be better if we copy the config folders to a tmp path, do required changes for the test there, so we can ensure that the tests do not conflict with each other. When two tests in parallel touch some files in the config folder this can result in race conditions. At the moment some tests touch files in the tests folder (e.g. grid file) but they are not conflicting. This could however happen easily in some future changes and would be annoying to debug.
- https://github.com/C2SM/Sirocco/pull/143#discussion_r2102129390
- https://github.com/C2SM/Sirocco/pull/143#discussion_r2102130356


src/sirocco/core/_tasks/icon_task.py
```diff
@@ -1,81 +1,65 @@
...
+        self._content = f90nml.read(path)
+        self._path = path
+
+    def update_from_specs(self, specs: dict[str, Any]) -> None:
+        """Updates the internal content from the specs."""
+        for section, params in specs.items():
+            section_name, k = self.section_index(section)
+            # Create section if non-existent
+            if section_name not in self.content:
+                # NOTE: f90nml will automatially create the corresponding nested f90nml.Namelist
```
agoscinski, line 47:
```suggestion
                # NOTE: f90nml will automatically create the corresponding nested f90nml.Namelist
```

src/sirocco/core/_tasks/icon_task.py
```diff
@@ -1,81 +1,65 @@
...
-            suffix = ("_".join([str(p) for p in self.coordinates.values()])).replace(" ", "_")
-            self.core_namelists[name].write(folder / (name + "_" + suffix), force=True)
 
-    def create_workflow_namelists(self, folder=None):
-        self.init_core_namelists()
-        self.update_core_namelists_from_config()
-        self.update_core_namelists_from_workflow()
-        self.dump_core_namelists(folder=folder)
+@dataclass(kw_only=True)
+class Namelist(models.ConfigNamelistSpec):
```
agoscinski, line 18: try to use `f90nml.Namelist`

tests/cases/large/data/ICON_namelists/icon_master.namelist_2025-01-01_00:00:00
```diff
@@ -1,6 +1,6 @@
 &master_nml
-    lrestart = .true.
-    read_restart_namelists = .true.
```
agoscinski, line 3: that was a bug in `is_restart`, not considering our port structure (having the port but with an empty list)

.github/workflows/ci.yaml
```diff
@@ -38,7 +38,7 @@ jobs:
       env:
         PYTEST_ADDOPTS: "--durations=0"
       run: |
-        hatch test -m "slow or not requires_icon" --parallel --cover
+        hatch test -m "not requires_icon" --parallel --cover
```
agoscinski, line 41: forgot to rebase
```suggestion
        hatch test -m "slow or not requires_icon" --parallel --cover
```

.github/workflows/test-integration.yml
```diff
@@ -58,3 +58,4 @@ jobs:
       run: |
         spack env activate .
         hatch test -m "requires_icon" --parallel tests
+
```
agoscinski, line 61: forgot to rebase
```suggestion
```

src/sirocco/core/_tasks/icon_task.py
```diff
@@ -64,7 +64,8 @@ def model_namelist(self) -> NamelistFile:
     @property
     def is_restart(self) -> bool:
         """Check if the icon task starts from the restart file."""
-        return self._AIIDA_ICON_RESTART_FILE_PORT_NAME in self.inputs
+        # restart_file must be present and nonempty
```
agoscinski, line 67:
```suggestion
        # restart port must be present and nonempty
```

tests/cases/small-shell/config/config.yml
```diff
@@ -50,7 +50,7 @@ data:
          src: data/input
      - initial_conditions:
          type: file
-         computer: localhost
+         #computer: localhost # TODO requires fix from #109
```
agoscinski, line 53:
```suggestion
         #computer: localhost # TODO requires fix from PR #136
```

GeigerJ2 (commented):
Thanks a lot for the work, @agoscinski! Went through everything once quickly and looks good. Will have another look, and then I think we can approve.

tests/cases/small-icon/config/ICON/model.namelist
```diff
@@ -0,0 +1,30 @@
+&run_nml
+ dtime                       = 1209600           ! time step of 300 seconds
```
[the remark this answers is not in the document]
agoscinski: ah sorry forgot to update this doc here
```suggestion
 dtime                       = 1209600           ! time step of 2 weeks
```

tests/cases/small-icon/config/ICON/icon_master.namelist
```diff
@@ -5,16 +5,13 @@
...
-&time_nml
- is_relative_time     = .false.
+ experimentStopDate = '2000-01-01T02:00:00Z'
+ restartTimeIntval    = 'PT1H'
+ checkpointTimeIntval = 'PT1H'
 /
 &master_model_nml
   model_name="atm"
-  model_namelist_filename="NAMELIST_exclaim_ape_R02B04"
+  model_namelist_filename="model.namelist"
```
leclairm, line 14: Why do you want to change this namelist filename? Is it hardcoded in the aiida-icon plugin? If yes, we need to still give the user the ability to name the file meaningfully.

tests/cases/small-icon/config/ICON/model.namelist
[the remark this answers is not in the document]
leclairm: It doesn't and should crash. This is 14 days in seconds. Don't know where it comes from.

tests/conftest.py
```diff
@@ -57,22 +59,40 @@ def icon_filepath_executable() -> str:
...
+            available=available_datas,
             generated=[models.ConfigGeneratedData(name="bar", type=models.DataType.DIR, src=pathlib.Path("bar"))],
         ),
         parameters={},
     )
 
 
 @pytest.fixture(scope="session")
-def minimal_invert_task_io_config() -> models.ConfigWorkflow:
+def minimal_invert_task_io_config(available_datas) -> models.ConfigWorkflow:
```
leclairm, line 95: I might be lost in pytest (again ...): what provides `available_datas`?

leclairm (commented):
Main point is about data location (not in config rootdir)

Side remark: For pyright to make sense of single dispatch functions, it needs `-> None` in the signature, otherwise, IDEs using it show errors (don't know how pythonic that is though)

src/sirocco/parsing/yaml_data_models.py
```diff
@@ -478,7 +483,7 @@ class DataType(enum.StrEnum):
 @dataclass(kw_only=True)
 class ConfigBaseDataSpecs:
     type: DataType
-    src: Path
+    src: Path | None = None
```
agoscinski, line 486: `src:` -> `bin:`

src/sirocco/core/graph_items.py
```diff
@@ -45,22 +45,60 @@ class Data(ConfigBaseDataSpecs, GraphItem):
...
+            config_rootdir=config_rootdir,
+            coordinates=coordinates,
+        )
+
+
+class AvailableData(Data):
+    src: Path
+
+    @classmethod
+    def from_config(cls, config: ConfigBaseData, config_rootdir: Path, coordinates: dict) -> Self:
```
agoscinski, line 64: validation can be now done in the yaml

src/sirocco/core/_tasks/icon_task.py
```diff
@@ -64,7 +64,8 @@ def model_namelist(self) -> NamelistFile:
...
+        return bool(self.inputs.get(self._AIIDA_ICON_RESTART_FILE_PORT_NAME, None))
```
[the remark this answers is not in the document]
agoscinski: done

src/sirocco/core/graph_items.py
```diff
@@ -45,22 +45,60 @@ class Data(ConfigBaseDataSpecs, GraphItem):
...
+            src=src,
             coordinates=coordinates,
         )
 
+    @staticmethod
+    def _validate_src(name: str, config_src: Path | None, config_rootdir: Path | None = None) -> Path | None:
+        if config_src is None:
+            return None
+        if config_rootdir is None and not config_src.is_absolute():
+            msg = f"Cannot specify relative path {config_src} for namelist while the rootdir is None"
```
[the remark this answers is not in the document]
agoscinski: usage of `config_rootdir` removed in https://github.com/C2SM/Sirocco/pull/153

tests/cases/small-icon/config/ICON/icon_master.namelist
agoscinski: I tried to be consistent with aiida-icon, there I guess rico just chose a name. it does not matter, one just needs to be consistent in the config.yml. I can change the name if you want.

src/sirocco/parsing/yaml_data_models.py
agoscinski: I put this also in a different PR to avoid noise

src/sirocco/parsing/yaml_data_models.py
```diff
@@ -513,7 +518,7 @@ class ConfigBaseData(_NamedBaseModel, ConfigBaseDataSpecs):
 
 
 class ConfigAvailableData(ConfigBaseData):
-    pass
+    src: Path
```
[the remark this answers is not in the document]
leclairm: I agree. Better than `src`. Fine for another PR.

src/sirocco/workgraph.py
```diff
@@ -267,23 +270,76 @@ def _create_shell_task_node(self, task: core.ShellTask):
...
+            raise ValueError(msg) from err
+
+        label_uuid = str(uuid.uuid4())
+        icon_code = aiida.orm.InstalledCode(
+            label=f"icon-{label_uuid}",
+            description="aiida_icon",
+            default_calc_job_plugin="icon.icon",
+            computer=computer,
+            filepath_executable=str(task.src),
+            with_mpi=False,
```
leclairm, line 291: What is `with_mpi` doing`? We'll obviously need mpi but it feels to me it should be rather handled by `aiida-icon`

tests/cases/large/data/ICON_namelists/icon_master.namelist_2025-01-01_00:00:00
leclairm: good catch, thanks.

tests/cases/small-icon/config/ICON/icon_master.namelist
leclairm: I would keep it as the name "exclaim_ape_R02B04" says something about the configuration.

leclairm (approved):
Fine for me. I have a few comments but not blocking. You can address them now if you want or we keep it for later.

src/sirocco/workgraph.py
agoscinski: In aiida the mpi option is part of the code configuration since for any code the way to run it with mpi is the same (at least this is the rational of the implementation in AiiDA, there are different mpi version AFAIK).

So for the CI in this PR, we don't want it because we compiled icor without mpi. Probably the easiest solution for early user milestone is to add opermpi to spack to make it true and always run CI. I opened a PR regarding this

Also, we made it an issue to make an option in the config yaml https://github.com/C2SM/Sirocco/issues/159

src/sirocco/workgraph.py
```diff
@@ -267,23 +270,76 @@ def _create_shell_task_node(self, task: core.ShellTask):
...
+            builder.model_namelist = aiida.orm.SinglefileData(buffer, task.model_namelist.name)
+
+        self._aiida_task_nodes[task_label] = self._workgraph.add_task(builder)
+
+    @functools.singledispatchmethod
+    def _link_output_node_to_task(self, task: core.Task, port: str, output: core.Data):  # noqa: ARG002
+        """Dispatch linking input to task based on task type."""
+
+        msg = f"method not implemented for task type {type(task)}"
+        raise NotImplementedError(msg)
```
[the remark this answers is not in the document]
agoscinski: Not sure what you mean here. Raising an error is the default now, what should be the default?

PR #150, opened 2025-05-09
independent namelistfile.py module

leclairm:


agoscinski (approved):
Looks good. One needs to just rename the function in the `conftest.py`. See type checker error. Already approving since minor change.

> tests/conftest.py:104: error: "IconTask" has no attribute "dump_icon_namelists"; maybe "dump_namelists"?  [attr-defined]

PR #151, opened 2025-05-16
Add minimal CLI interface using typer and rich

GeigerJ2:


GeigerJ2:
Note: People will need simple way to access current run directory, wrap `verdi` commands, or provide minimal `sirocco` wrapper for that

PR #153, opened 2025-05-28
Enforce usage of absolute path and computer

agoscinski:
Everywhere instead of GeneratedData. In AvailableData I also did not add it since we require #136 to be merged before we can use `RemoteData` in shell tasks. I consequently added it as TODO.

To make it clear to the user that the scripts and icon executable are on a computer, I made computer a required argument (I would like to do the same for data but need #136). However, in the parsing process we do not have a mechanism to pass default arguments from the root task to the other tasks so I need to add everywhere `computer`. Since the root task feature is not essential for a minimum, I marked this with a TODO referencing one of the issues about root.

When defining an icon task in the config yml, it is a bit opaque that the computer is not applied on the namelists but only on the executable path (`src`). Bundling the computer and the executable path to a subsection is not a solution since we want to also pass submission information over the computer that can be overwritten on the same indentation level. I therefore only allow relative paths for namelists to make clear that they are an exception.

I realized a design flaw in the usage of environment variables. I wanted to use environment variables to use absolute paths without hardcoding the repos directory (as this is different for every dev/CI). The resolvement of the environment variables should mainly happen during the running time of the workflow since the environment variables could be specified on a remote machine. Now aiida does not resolve the environment variables when linking them to the aiida working directory (I don't know if there is a way but since this logic it is not trivial to fix), so the environment variable is escaped. There needs to be a step that resolves the escaped environment variables. Right now I add a script that does this as prepend text to the computer (see conftest.py in PR 153). I think we need the environment variables when we start to use the CSCS CI as we might want to run the same workflow but with a different grid or namelist parameter. It is hard to say for sure now if it is really needed for the CSCS CI, since we could make another dedicated use case for this, but it might be needed.

### DIFF
--- a/src/sirocco/core/_tasks/icon_task.py
+++ b/src/sirocco/core/_tasks/icon_task.py
@@ -95,7 +95,7 @@ class IconTask(models.ConfigIconTaskSpecs, Task):
             namelist.dump(directory / filename)
 
     @classmethod
-    def build_from_config(cls: type[Self], config: models.ConfigTask, **kwargs: Any) -> Self:
+    def build_from_config(cls: type[Self], config: models.ConfigTask, config_rootdir: Path, **kwargs: Any) -> Self:
         config_kwargs = dict(config)
         del config_kwargs["parameters"]
         # The following check is here for type checkers.
@@ -105,10 +105,12 @@ class IconTask(models.ConfigIconTaskSpecs, Task):
             raise TypeError
 
         config_kwargs["namelists"] = [
-            NamelistFile.from_config(config=config_namelist) for config_namelist in config_kwargs["namelists"]
+            NamelistFile.from_config(config=config_namelist, config_rootdir=config_rootdir)
+            for config_namelist in config_kwargs["namelists"]
         ]
 
         self = cls(
+            config_rootdir=config_rootdir,
             **kwargs,
             **config_kwargs,
         )
--- a/src/sirocco/core/_tasks/icon_task.py
+++ b/src/sirocco/core/_tasks/icon_task.py
@@ -32,7 +32,6 @@ class IconTask(models.ConfigIconTaskSpecs, Task):
             msg = f"Failed to read master namelists. Could not find {self._MASTER_NAMELIST_NAME!r} in namelists {self.namelists}"
             raise ValueError(msg)
         self._master_namelist = master_namelist
-        self.src = self._validate_src(self.src, self.config_rootdir)
 
         # retrieve model namelist name from master namelist
         if (master_model_nml := self._master_namelist.namelist.get(self._MASTER_MODEL_NML_SECTION, None)) is None:
@@ -106,8 +105,7 @@ class IconTask(models.ConfigIconTaskSpecs, Task):
             raise TypeError
 
         config_kwargs["namelists"] = [
-            NamelistFile.from_config(config=config_namelist, config_rootdir=kwargs["config_rootdir"])
-            for config_namelist in config_kwargs["namelists"]
+            NamelistFile.from_config(config=config_namelist) for config_namelist in config_kwargs["namelists"]
         ]
 
         self = cls(
@@ -116,18 +114,3 @@ class IconTask(models.ConfigIconTaskSpecs, Task):
         )
         self.update_icon_namelists_from_workflow()
         return self
-
-    @staticmethod
-    def _validate_src(config_src: Path, config_rootdir: Path | None = None) -> Path:
-        if config_rootdir is None and not config_src.is_absolute():
-            msg = f"Cannot specify relative path {config_src} for namelist while the rootdir is None"
-            raise ValueError(msg)
-
-        src = config_src if config_rootdir is None else (config_rootdir / config_src)
-        if not src.exists():
-            msg = f"Icon executable in path {src} does not exist."
-            raise FileNotFoundError(msg)
-        if not src.is_file():
-            msg = f"Icon executable in path {src} is not a file."
-            raise OSError(msg)
-        return src
--- a/src/sirocco/core/graph_items.py
+++ b/src/sirocco/core/graph_items.py
@@ -140,10 +140,10 @@ class Task(ConfigBaseTaskSpecs, GraphItem):
         return new
 
     @classmethod
-    def build_from_config(cls: type[Self], config: ConfigTask, **kwargs: Any) -> Self:
+    def build_from_config(cls: type[Self], config: ConfigTask, config_rootdir: Path, **kwargs: Any) -> Self:
         config_kwargs = dict(config)
         del config_kwargs["parameters"]
-        return cls(**kwargs, **config_kwargs)
+        return cls(config_rootdir=config_rootdir, **kwargs, **config_kwargs)
 
     def link_wait_on_tasks(self, taskstore: Store[Task]) -> None:
         self.wait_on = list(
--- a/src/sirocco/core/graph_items.py
+++ b/src/sirocco/core/graph_items.py
@@ -45,53 +45,19 @@ class Data(ConfigBaseDataSpecs, GraphItem):
     color: ClassVar[str] = field(default="light_blue", repr=False)
 
     @classmethod
-    def from_config(
-        cls, config: ConfigBaseData, config_rootdir: Path, coordinates: dict
-    ) -> AvailableData | GeneratedData:
+    def from_config(cls, config: ConfigBaseData, coordinates: dict) -> AvailableData | GeneratedData:
         data_class = AvailableData if isinstance(config, ConfigAvailableData) else GeneratedData
-
-        return data_class.from_config(
-            config=config,
-            config_rootdir=config_rootdir,
-            coordinates=coordinates,
-        )
+        config_kwargs = dict(config)
+        del config_kwargs["parameters"]
+        return data_class(coordinates=coordinates, **config_kwargs)
 
 
 class AvailableData(Data):
     src: Path
 
-    @classmethod
-    def from_config(cls, config: ConfigBaseData, config_rootdir: Path, coordinates: dict) -> Self:
-        src = cls._validate_src(config.src, config_rootdir)
-        return cls(
-            name=config.name,
-            computer=config.computer,
-            type=config.type,
-            src=src,
-            coordinates=coordinates,
-        )
-
-    @staticmethod
-    def _validate_src(config_src: Path | None, config_rootdir: Path | None = None) -> Path | None:
-        if config_src is None:
-            return None
-        if config_rootdir is None and not config_src.is_absolute():
-            msg = f"Cannot specify relative path {config_src} for namelist while the rootdir is None"
-            raise ValueError(msg)
-
-        return config_src if config_rootdir is None else (config_rootdir / config_src)
-
 
 class GeneratedData(Data):
-    @classmethod
-    def from_config(cls, config: ConfigBaseData, config_rootdir: Path, coordinates: dict) -> Self:  # noqa: ARG003 # we need to keep same signature as for AvailableData
-        return cls(
-            name=config.name,
-            computer=config.computer,
-            type=config.type,
-            src=config.src,
-            coordinates=coordinates,
-        )
+    pass
 
 
 @dataclass(kw_only=True)
--- a/src/sirocco/core/namelistfile.py
+++ b/src/sirocco/core/namelistfile.py
@@ -1,3 +1,4 @@
+import os
 import re
 from dataclasses import dataclass, field
 from pathlib import Path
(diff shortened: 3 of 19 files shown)


.github/workflows/test-integration.yml
```diff
@@ -52,9 +52,15 @@ jobs:
...
       env:
         PYTEST_ADDOPTS: "--durations=0"
       run: |
         spack env activate .
         hatch test -m "requires_icon" --parallel tests
+      continue-on-error: true
+
+    - name: Setup upterm session
+      if: steps.testrun.outcome == 'failure'
+      uses: lhotari/action-upterm@v1
```
agoscinski: This can be removed. was just for debugging useful

src/sirocco/core/_tasks/icon_task.py
```diff
@@ -106,28 +105,14 @@ def build_from_config(cls: type[Self], config: models.ConfigTask, **kwargs: Any)
...
-            raise ValueError(msg)
-
-        src = config_src if config_rootdir is None else (config_rootdir / config_src)
-        if not src.exists():
-            msg = f"Icon executable in path {src} does not exist."
-            raise FileNotFoundError(msg)
-        if not src.is_file():
-            msg = f"Icon executable in path {src} is not a file."
-            raise OSError(msg)
-        return src
```
agoscinski, line 133: the icon bin cannot be validated anymore since it can be remote. Has to happen during runtime of workflow

src/sirocco/workgraph.py
```diff
@@ -190,10 +190,10 @@ def _add_aiida_input_data_node(self, data: core.AvailableData):
...
                 remote_path=str(data.src), label=label, computer=computer
             )
-        elif data.type == "file":
-            self._aiida_data_nodes[label] = aiida.orm.SinglefileData(label=label, file=data_full_path)
-        elif data.type == "dir":
-            self._aiida_data_nodes[label] = aiida.orm.FolderData(label=label, tree=data_full_path)
+        elif data.computer is None and data.type == "file":
+            self._aiida_data_nodes[label] = aiida.orm.SinglefileData(label=label, file=os.path.expandvars(data.src))
+        elif data.computer is None and data.type == "dir":
+            self._aiida_data_nodes[label] = aiida.orm.FolderData(label=label, tree=os.path.expandvars(data.src))
```
agoscinski, line 196: just to be more explicit but this should not change logic

src/sirocco/core/graph_items.py
```diff
@@ -46,52 +46,20 @@ class Data(ConfigBaseDataSpecs, GraphItem):
...
         data_class = AvailableData if isinstance(config, ConfigAvailableData) else GeneratedData
-
-        return data_class.from_config(
-            config=config,
-            config_rootdir=config_rootdir,
-            coordinates=coordinates,
-        )
+        config_kwargs = dict(config)
+        del config_kwargs["parameters"]
+        return data_class(coordinates=coordinates, **config_kwargs)
```
agoscinski, line 54: To pass computer argument correctly. Not sure how it worked before since computer was not passed ->  data.compute is None -> in aiida workgraph creation it creates SinglefileData -> aiida-icon complains because it was RemoteData

src/sirocco/workgraph.py
```diff
@@ -221,18 +221,25 @@ def _create_shell_task_node(self, task: core.ShellTask):
             if task.src is None:
                 msg = "src must be specified when command path is relative"
                 raise ValueError(msg)
-            command = str((task.config_rootdir / task.src).parent / cmd_path)
+            command = str(task.src.parent / cmd_path)
+
+        from aiida_shell import ShellCode
+
+        label_uuid = str(uuid.uuid4())
+        # PRCOMMENT aiida-shell created the code before from the command we passed but this only works if it can resolve commands. Since we now want to support environment variables that possibly cannot be resoved until submission (because it uses environment variables in one of the source files)
```
agoscinski, line 229: read this comment please

tests/conftest.py
```diff
@@ -163,10 +164,54 @@ def serialize_nml(config_paths: dict[str, pathlib.Path], workflow: workflow.Work
...
+rm "$link"
+ln -s "$resolved" "$link"
+else
+echo " !! Expanded path does not exist: $resolved"
+fi
+else
+echo " -> No environment variable to expand."
+fi
+fi
+done"""
```
agoscinski, line 216: it is not tabbed properly because aiida has a bug to remove everything beyond 80 character in the lin. Need to make an issue in aiida-core.

src/sirocco/core/graph_items.py
agoscinski: Yes kind of the same fix just passing `kwargs`. I was not sure how PR #143 (on which this PR is based on) was working before. But I see now that the computer was passed later in the `AvailableData` class. This PR removed this constructor so had to bring back passing the computer here. Actually, it does not make so much sense to have `computer` for `GeneratedData` need to fix this.

tests/cases/small-icon/config/ICON/bin/icon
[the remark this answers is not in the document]
agoscinski: Yes I am also confused. I cannot remove it. Will try to add empty file again

tests/conftest.py
agoscinski: Good idea!

src/sirocco/workgraph.py
GeigerJ2: IIRC we agreed that the user needs to take care of that via the script they provide, which will eventually become the `code` executed by `aiida-shell`. I'd need to see an actual use case of which env variables would be used for what and where they would be sourced.

I think we have to anyway add our own custom functionality for code creation based on the config, even before `aiida-shell`, as I see it too restrictive, as well. `ShellCode` inherits from `InstalledCode`, so even `PortableCode`, which I think would also be a common use case for us, doesn't seem to supported? In the case of `PortableCode`, that could be a local script contained together with the sirocco config folder, that could possibly also contain files with configuration settings for that script, e.g., that contain environment variables, etc.

src/sirocco/core/namelistfile.py
```diff
@@ -60,6 +60,8 @@ def dump(self, path: Path) -> None:
 
     @staticmethod
     def _validate_namelist_path(config_namelist_path: Path, config_rootdir: Path | None = None) -> Path:
+        if config_namelist_path.is_absolute():
+            msg = f"Namelist path {config_namelist_path} must be relative with respect to config file."
         if config_rootdir is None and not config_namelist_path.is_absolute():
             msg = f"Cannot specify relative path {config_namelist_path} for namelist while the rootdir is None"
```
leclairm, line 66: That should be removed.

src/sirocco/parsing/yaml_data_models.py
```diff
@@ -250,7 +250,7 @@ class ConfigBaseTaskSpecs:
     Any of these keys can be None, in which case they are inherited from the root task.
     """
 
-    computer: str | None = None
+    computer: str
```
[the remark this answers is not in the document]
leclairm: > I'm fine with enforcing this (rather than, e.g., assuming `localhost` if None), even though it's more verbose, but at least it doesn't lead to unexpected consequences down the line.

Agreed. Also goes the same direction as we said about testing: even if running on localhost, make as if it were a remote computer.

src/sirocco/workgraph.py
leclairm: Sorry for being late. Yes, this is the design we wanted: users provide source files if needed (the command could also point to an executable available on the remote at runtime). They are specified in the `src` entry of the config file (can be a directory or a single file) and copied over by aiida to where the workflow runs. If environment variables are needed, they are set there.

Typical use case. A config file like this:
```yaml
[...]
tasks:
  - postproc:
      plugin: shell
      src: postproc/scripts
      command: "./postproc/scripts/main_script.sh --input {PORT::streams}"
[...]
```
The content of the config rootdir:
```shell
❯ tree
.
└── postproc
    └── scripts
        ├── env_var.sh
        └── main_script.sh
```
and in `scripts/main_script.sh` you have stg like `source env_vars.sh`.
There's actually a key piece of information here for the users to document: when pointing to other source files within your scripts, beware that the origin working directory is not the location of the script (`main_script.sh` in out case)

tests/cases/small-icon/config/config.yml
```diff
@@ -41,36 +41,37 @@ cycles:
...
          computer: localhost
      - ECHAM6_CldOptProps:
          type: file
-         src: ICON/ECHAM6_CldOptProps.nc
+         src: $TEST_ROOTDIR/tests/cases/small-icon/config/ICON/ECHAM6_CldOptProps.nc
          computer: localhost
      - rrtmg_sw:
          type: file
-         src: ICON/rrtmg_sw.nc
+         src: $TEST_ROOTDIR/tests/cases/small-icon/config/ICON/rrtmg_sw.nc
```
agoscinski, line 70: shell task is always relative also, does not provide computer argument, when renaming for icon task `src` -> `bin` it should be clear to the user that this is differently handled

agoscinski:
For future PR:
- Shell task `src` should always be relative, and not support computer argument. When renaming for `src` for icon task to `bin` (open PR TODO) it should be clear to the user that this is differently handled.

src/sirocco/workgraph.py
GeigerJ2: I'm still not sure why we need to append the UUID now to every code label, such that we end up with this:
```shell
❯ verdi process list -ap 1
  PK  Created    Process label                                                                                                                                                           ♻    Process State    Process status
----  ---------  ----------------------------------------------------------------------------------------------------------------------------------------------------------------------  ---  ---------------  ----------------
 214  2m ago     WorkGraph<config>                                                                                                                                                            ⏹ Finished [0]
 217  2m ago     ShellJob</home/geiger_j/aiida_projects/swiss-twins/git-repos/Sirocco/tests/cases/small-shell/config/scripts/cleanup.py-1090a7b0-b956-4c05-beac-557a195562d3@localhost>       ⏹ Finished [0]
 221  2m ago     ShellJob</home/geiger_j/aiida_projects/swiss-twins/git-repos/Sirocco/tests/cases/small-shell/config/scripts/icon.py-413689db-e7ad-40b6-a4ac-0f186837a0dc@localhost>          ⏹ Finished [0]
 235  2m ago     ShellJob</home/geiger_j/aiida_projects/swiss-twins/git-repos/Sirocco/tests/cases/small-shell/config/scripts/icon.py-66aa102e-4d8b-4e5d-abe1-44a2c2613ca6@localhost>          ⏹ Finished [0]
 245  2m ago     ShellJob</home/geiger_j/aiida_projects/swiss-twins/git-repos/Sirocco/tests/cases/small-shell/config/scripts/icon.py-5deff559-5262-4f47-af0d-93ab7efdf456@localhost>          ⏹ Finished [0]

❯ verdi code list
Full label                                                                                                                                                      Pk  Entry point
------------------------------------------------------------------------------------------------------------------------------------------------------------  ----  -------------------------
/home/geiger_j/aiida_projects/swiss-twins/git-repos/Sirocco/tests/cases/small-shell/config/scripts/icon.py-413689db-e7ad-40b6-a4ac-0f186837a0dc@localhost      208  core.code.installed.shell
/home/geiger_j/aiida_projects/swiss-twins/git-repos/Sirocco/tests/cases/small-shell/config/scripts/icon.py-66aa102e-4d8b-4e5d-abe1-44a2c2613ca6@localhost      209  core.code.installed.shell
/home/geiger_j/aiida_projects/swiss-twins/git-repos/Sirocco/tests/cases/small-shell/config/scripts/icon.py-5deff559-5262-4f47-af0d-93ab7efdf456@localhost      210  core.code.installed.shell
/home/geiger_j/aiida_projects/swiss-twins/git-repos/Sirocco/tests/cases/small-shell/config/scripts/cleanup.py-1090a7b0-b956-4c05-beac-557a195562d3@localhost   211  core.code.installed.shell
```

Could you elaborate again?

src/sirocco/workgraph.py
```diff
@@ -221,18 +221,25 @@ def _create_shell_task_node(self, task: core.ShellTask):
...
         metadata: dict[str, Any] = {}
-        ## Source file
-        env_source_paths = [
-            env_source_path
-            if (env_source_path := Path(env_source_file)).is_absolute()
-            else (task.config_rootdir / env_source_path)
-            for env_source_file in task.env_source_files
-        ]
+        # Files that are sourced before the execution of the script
+        env_source_paths = [Path(env_source_file) for env_source_file in task.env_source_files]
```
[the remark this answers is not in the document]
GeigerJ2: I'd still keep the code in this PR, though, and remove it in a separate one to keep the history more clean

tests/cases/large/config/config.yml
```diff
@@ -150,27 +156,29 @@ tasks:
...
+      src: $TEST_ROOTDIR/tests/cases/large/config/scripts/scripts/post_clean.sh
       command: "post_clean.sh --archive {PORT::archive} --clean {PORT::clean}"
       nodes: 1
       walltime: 00:01:00
 data:
   available:
     - grid_file:
         type: file
-        src: data/grid
+        src: $TEST_ROOTDIR/tests/cases/large/config/data/grid
```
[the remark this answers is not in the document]
agoscinski: now issue https://github.com/C2SM/Sirocco/issues/170

PR #155, opened 2025-06-03
Add `from_yaml_str` method to config workflow

GeigerJ2:
Simple utility method to avoid having to write temporary files and then reading them in again. Also allows using placeholders in the string that can be resolved at runtime, e.g., to paths.


GeigerJ2:
Implementation basically the same as `validate_yaml_content` function (with some additional shortcuts, namely setting `name` and `rootdir` automatically if not provided), and verification (that the YAML string is not empty). Don't add for now, to avoid code duplication and duplicated maintenance cost, but use `validate_yaml_content` function instead.

PR #156, opened 2025-06-03
Improve error handling when secrets are not accessible

agoscinski:
When a PR is merged from a fork it does not have access to the upstream secrets needed for spack buildcache. Since the error in this case is very opaque we check the condition earlier in the integration workflow and error out.


GeigerJ2 (approved):
LGTM!

(any way one could get the secrets, though? 👀)

PR #157, opened 2025-06-03
Fix remote data copying/symlinking with dates and parameters

GeigerJ2:


pyproject.toml
```diff
@@ -75,6 +76,14 @@ ignore = [
...
 ]
 
+[tool.ruff.lint.per-file-ignores]
+"tests/*py" = [
+    "SLF001",  # Private member accessed
+    "S101",    # Use of assert detected
+    "T201",    # `print` found
+    "PLR2004", # Magic value used in comparison
+]
+
```
agoscinski, line 86: Would not introduce ignores for all cases. Also it seems related your local config. In the CI the hatch fmt pass without the `noqa` introduced in this PR. Could it be that you use py3.13?
```suggestion
```

tests/conftest.py
```diff
@@ -48,7 +48,7 @@ def icon_grid_simple_path(pytestconfig):
 
 @pytest.fixture
 def icon_filepath_executable() -> str:
-    which_icon = subprocess.run(["which", "icon"], capture_output=True, check=False)
+    which_icon = subprocess.run(["which", "icon"], capture_output=True, check=False)  # noqa: S607
```
agoscinski, line 51: I think not needed, need to check why you have this issues.
```suggestion
    which_icon = subprocess.run(["which", "icon"], capture_output=True, check=False)
```

tests/conftest.py
```diff
@@ -154,7 +158,7 @@ def serialize_nml(config_paths: dict[str, pathlib.Path], workflow: workflow.Work
 
 def pytest_configure(config):
     if config.getoption("reserialize"):
-        print("Regenerating serialized references")  # noqa: T201 # this is actual UX, not a debug print
+        print("Regenerating serialized references")  # this is actual UX, not a debug print
```
agoscinski, line 161: changed to logging in PR #160

agoscinski (changes requested):

tests/test_workgraph.py
```diff
@@ -0,0 +1,727 @@
...
+            coordinates={},
+        ),
+    ]
+
+    # Test that conflict detection works
+    output_conflicts = [inp for inp in inputs if inp.name == "output"]
+    other_conflicts = [inp for inp in inputs if inp.name == "other"]
+
+    assert len(output_conflicts) == 2  # Should need unique filenames
+    assert len(other_conflicts) == 1  # Should use simple filename
```
agoscinski, line 73: What logic is this testing? I only see the usage of `GeneratedData` but I don't see that any specific logic of `GeneratedData` is used.

pyproject.toml
GeigerJ2: All of these issues appeared for me during `hatch fmt` on the testing files, so I was also confused. Let's see what happens when we remove. I use Python 3.12.7.

tests/conftest.py
```diff
@@ -21,7 +21,7 @@ def __init__(self, url: str, response: requests.Response):
 
 
 def download_file(url: str, file_path: pathlib.Path):
-    response = requests.get(url)
+    response = requests.get(url)  # noqa: S113 request-without-timeout
```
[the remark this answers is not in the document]
GeigerJ2: Weird, having this locally with Python 3.12.7...

tests/conftest.py
```diff
@@ -87,7 +87,7 @@ def minimal_invert_task_io_config() -> models.ConfigWorkflow:
                     ),
                     models.ConfigCycleTask(
                         name="task_a",
-                        inputs=[models.ConfigCycleTaskInput(name="availalble", port="None")],
+                        inputs=[models.ConfigCycleTaskInput(name="available", port="None")],
```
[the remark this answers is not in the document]
GeigerJ2: Approved!

pyproject.toml
```diff
@@ -75,6 +76,7 @@ ignore = [
   "TRY003", # write custom error messages for formatting
 ]
 
+[tool.ruff.lint.per-file-ignores]
```
agoscinski, line 79:
```suggestion
```

src/sirocco/workgraph.py
```diff
@@ -293,21 +311,67 @@ def _link_wait_on_to_task(self, task: core.Task):
...
+                # Count how many inputs have the same base name
+                same_name_count = sum(1 for inp in task.input_data_nodes() if inp.name == input_.name)
+
+                if same_name_count > 1:
+                    # Multiple data nodes with same base name - use full label as filename
+                    # to ensure uniqueness in working directory
+                    filename = input_label
+                else:
+                    # Single data node with this name - can use simple filename
+                    filename = Path(input_.src).name if hasattr(input_, "src") else input_.name
```
agoscinski, line 368: Maybe its simpler to just always use the input_label with the coordinates as filename so it is more predictable? We can merge it like it is now but you have any thoughts on always using the if-case?

GeigerJ2: Yeah, fair point. I feel like that's a trade-off/preference, between simplicity and predictability. The latter approach leads to always the same `filename` structure, but these could be unnecessarily lengthy ones. Will add a comment.

agoscinski (approved):
Thanks for the work on this nasty issue!

PR #164, opened 2025-06-04
Require `computer` for available data and remove `type`

agoscinski:
With the support of `aiida.orm.RemoteData` we can remove the `type` specification for data as both is supported. For a config file that is more transparent about where the data is located we require to specify the `computer`.

### DIFF
--- a/src/sirocco/parsing/yaml_data_models.py
+++ b/src/sirocco/parsing/yaml_data_models.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import enum
 import itertools
 import re
 import time
@@ -486,17 +485,10 @@ class ConfigIconTask(ConfigBaseTask, ConfigIconTaskSpecs):
         return nmls
 
 
-class DataType(enum.StrEnum):
-    FILE = enum.auto()
-    DIR = enum.auto()
-
-
 @dataclass(kw_only=True)
 class ConfigBaseDataSpecs:
-    type: DataType
     src: Path | None = None
     format: str | None = None
-    computer: str | None = None
 
 
 class ConfigBaseData(_NamedBaseModel, ConfigBaseDataSpecs):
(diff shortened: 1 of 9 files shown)


pyproject.toml
```diff
@@ -95,7 +96,7 @@ extra-dependencies = [
     "ipdb"
 ]
 default-args = []
-extra-args = ["--doctest-modules", '-m not slow', '-m not requires_icon']
+extra-args = ["--doctest-modules", "-m", "not slow and not requires_icon"]
```
agoscinski, line 99: not sure why this has been changed. will try to move to different PR

pyproject.toml
```diff
@@ -135,7 +136,7 @@ extra-dependencies = [
 ]
 
 [tool.hatch.envs.types.scripts]
-check = "mypy --no-incremental {args:.}"
+check = "mypy --exclude 'tests/cases/*' --no-incremental {args:.}"
```
agoscinski, line 139: will try to move to different PR

src/sirocco/core/_tasks/icon_task.py
```diff
@@ -22,7 +22,6 @@ class IconTask(models.ConfigIconTaskSpecs, Task):
 
     def __post_init__(self):
         super().__post_init__()
-
```
agoscinski, line 25: not sure if fmt, will try to undo

src/sirocco/core/_tasks/icon_task.py
```diff
@@ -64,7 +63,8 @@ def model_namelist(self) -> NamelistFile:
     @property
     def is_restart(self) -> bool:
         """Check if the icon task starts from the restart file."""
-        return self._AIIDA_ICON_RESTART_FILE_PORT_NAME in self.inputs
+        # restart port must be present and nonempty
```
agoscinski, line 66: will undo, should be in different PR

src/sirocco/workgraph.py
```diff
@@ -190,22 +189,13 @@ def _add_aiida_input_data_node(self, data: core.AvailableData):
...
-        else:
-            msg = f"Data type {data.type!r} not supported. Please use 'file' or 'dir'."
-            raise ValueError(msg)
+        try:
+            computer = aiida.orm.load_computer(data.computer)
+        except NotExistent as err:
+            msg = f"Could not find computer {data.computer!r} for input {data}."
+            raise ValueError(msg) from err
+        # `remote_path` must be str not PosixPath to be JSON-serializable
+        self._aiida_data_nodes[label] = aiida.orm.RemoteData(remote_path=str(data.src), label=label, computer=computer)
```
GeigerJ2, line 198: Should we capture `data.src` being wrong/non-existent here, which would except `orm.RemoteData` creation?

src/sirocco/parsing/yaml_data_models.py
```diff
@@ -508,35 +500,28 @@ class ConfigBaseData(_NamedBaseModel, ConfigBaseDataSpecs):
...
 
-class ConfigGeneratedData(ConfigBaseData):
-    @field_validator("computer")
-    @classmethod
-    def invalid_field(cls, value: str | None) -> str | None:
-        if value is not None:
-            msg = "The field 'computer' can only be specified for available data."
-            raise ValueError(msg)
-        return value
+class ConfigGeneratedData(ConfigBaseData): ...
```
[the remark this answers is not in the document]
GeigerJ2: Note: `ConfigBaseDataSpecs` has `src`, so my comment is void. Overlooked this.

src/sirocco/workgraph.py
agoscinski: okay using aiida transport we can also check remote

PR #168, opened 2025-06-06
Default to run icon with mpi

agoscinski:
For the initial tests we installed `icon` without mpi support. For the early user milestone we want to run `icon` however with mpi support. So we hardcode to use icon with mpi and install openmpi in the CI.

I remember now the original problem I did not include mpi. It was because I was not sure how to pass `tot_num_mpiprocs`. Currently in this PR overwrite the mpi command, but for production we need to do pass the number of mpi procs differently. Need to check real submission script.


.github/workflows/test-integration.yml
```diff
@@ -62,4 +62,8 @@ jobs:
         PYTEST_ADDOPTS: "--durations=0"
       run: |
         spack env activate .
-        hatch test -m "requires_icon" --parallel tests
+        hatch test -m "requires_icon" --parallel -- -s tests 
+
+    - name: Setup upterm session
+      uses: lhotari/action-upterm@v1
+      if: failure()
```
agoscinski, line 69:
```suggestion
        hatch test -m "requires_icon" --parallel tests 

    - name: Setup upterm session
      uses: lhotari/action-upterm@v1
      if: failure()
```

tests/conftest.py
```diff
@@ -215,3 +215,5 @@ def configure_aiida_localhost(test_rootdir, aiida_localhost):
     fi
 done"""
     aiida_localhost.set_prepend_text(prepend_text)
+    aiida_localhost.set_mpirun_command(['mpirun', '-np', '1'])
```
agoscinski, line 218: @GeigerJ2 how one sets again `tot_num_mpiprocs`? for the test it works but I think aiida had a proper solution that also considered the number of nodes

tests/test_wc_workflow.py
```diff
@@ -82,6 +82,15 @@ def test_run_workgraph_with_icon(icon_filepath_executable, config_paths, tmp_pat
...
     output_node = aiida_workflow.run()
+    if not output_node.is_finished_ok:
+        from aiida.cmdline.utils.common import get_calcjob_report, get_workchain_report
+        # overall report but often not enough to really find the bug, one has to go to calcjob
+        print(get_workchain_report(output_node, levelname='REPORT'))
+        # the calcjobs are typically stored in 'called_descendants'
+        for node in output_node.called_descendants:
+            print(f"{node.process_label} workdir:", node.get_remote_workdir())
+            print(f"{node.process_label} report:\n", get_calcjob_report(node))
+
```
agoscinski, line 93: remove debugging
```suggestion
```

.github/workflows/test-integration.yml
```diff
@@ -62,4 +62,4 @@ jobs:
         PYTEST_ADDOPTS: "--durations=0"
       run: |
         spack env activate .
-        hatch test -m "requires_icon" --parallel tests
+        hatch test -m "requires_icon" --parallel tests
```
agoscinski, line 65:
```suggestion
        hatch test -m "requires_icon" --parallel tests
```

tests/conftest.py
agoscinski: I removed to make my life easier

GeigerJ2:
Currently, passing the resource parameters from the config YAML to AiiDA does work, when using the `SlurmScheduler` (rather than `DirectScheduler` that would be used for `localhost`), so that's "solved".  However, to properly run the integration tests, we need a working SLURM setup on the GHA runner.  It seems like the CI manages to set this up:

```shell
Testing SLURM...
PARTITION AVAIL  TIMELIMIT  NODES  STATE NODELIST
debug*       up   infinite      1   idle fv-az1914-166
             JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
```

However, the workflow still doesn't actually run through:

```shell
6/11/2025 04:15:43 PM <7181> aiida.orm.nodes.process.workflow.workchain.WorkChainNode: [REPORT] [17|WorkGraphEngine|continue_workgraph]: tasks ready to run: calcjob1,cleanup
06/11/2025 04:15:43 PM <7181> aiida.orm.nodes.process.workflow.workchain.WorkChainNode: [REPORT] [17|WorkGraphEngine|on_wait]: Process status: Waiting for child processes: 18, 21
Error: The operation was canceled.
```

Unlike in #183, in this setup, we wouldn't be properly testing true remote files, as we're still on the localhost GHA runner, so the files from our `config` directory in this repository are globally available.

GeigerJ2:
Issue warning and ignore MPI options when transport is not firecrest (for now), to make all CI pass

agoscinski:
Has been solved by PR #190

PR #172, opened 2025-06-10
Enforce `ShellTask.src` to be relative path

agoscinski:


src/sirocco/core/_tasks/shell_task.py
```diff
@@ -1,10 +1,45 @@
...
+
+if TYPE_CHECKING:
+    from pathlib import Path
 
 
 @dataclass(kw_only=True)
-class ShellTask(ConfigShellTaskSpecs, Task): ...
+class ShellTask(models.ConfigShellTaskSpecs, Task):
+    @classmethod
+    def build_from_config(cls: type[Self], config: models.ConfigTask, config_rootdir: Path, **kwargs: Any) -> Self:
```
agoscinski, line 16: this function is mostly a copy past from IconTask. Only change the validation

src/sirocco/parsing/yaml_data_models.py
```diff
@@ -283,13 +283,8 @@ class ConfigShellTaskSpecs:
...
         default=None,
-        metadata={
-            "description": (
-                "If `src` not absolute, this ends up to be relative to the root directory of the config file."
-                "This should also be solved by registering `Code`s in AiiDA for the required scripts."
-                "See issues #60 and #127."
-            )
-        },
+        metadata={"description": ("Script file relative to the config directory.")},
+        repr=False
```
agoscinski, line 287: This is needed, because we now have in the repr the real absolute path (no environment variable), this is similar as for IconTask.src/bin

tests/cases/large/data/config.txt
```diff
@@ -14,7 +14,6 @@ cycles:
             walltime: time.struct_time(tm_year=1900, tm_mon=1, tm_mday=1, tm_hour=0, tm_min=2, tm_sec=0, tm_wday=0, tm_yday=1, tm_isdst=-1)
             cycle point: []
             plugin: 'shell'
-            src: scripts/extpar
```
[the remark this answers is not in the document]
agoscinski: So because it is a relative path now, we need to append the config rootdir in core. Then this full path becomes part of the representation so I removed it (the same for namelists paths). In the serialization it is a bit hard to get the relative path back as we do not store the information. We do not have the information because the resolvement of  the relative path only happens in one constructor, the `__init__` constructor does not even know about the `config_rootdir`. For absolute paths this mechanism does not exist and we use the `/TESTS_ROOTDIR` tag which is not depending on the developer environment. So I don't have a good idea how to simply solve this issue with the relative paths so I removed them from the repr for now. We did this in the past also for the `IconTask.bin` which was relative at some point.

src/sirocco/core/_tasks/shell_task.py
```diff
@@ -1,10 +1,45 @@
...
+    def _validate_src(config_src: Path, config_rootdir: Path) -> Path:
+        if config_src.is_absolute():
+            msg = f"Namelist path {config_src} must be relative with respect to config file."
+        src = config_src if config_rootdir is None else (config_rootdir / config_src)
+        if not src.exists():
+            msg = f"Namelist in path {src} does not exist."
+            raise FileNotFoundError(msg)
+        if not src.is_file():
+            msg = f"Namelist in path {src} is not a file."
+            raise OSError(msg)
```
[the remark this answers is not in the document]
agoscinski: thanks for the catch

src/sirocco/core/_tasks/shell_task.py
```diff
@@ -1,10 +1,45 @@
...
+            **config_kwargs,
+        )
+        if self.src is not None:
+            self.src = self._validate_src(self.src, config_rootdir)
+        return self
+
+    @staticmethod
+    def _validate_src(config_src: Path, config_rootdir: Path) -> Path:
+        if config_src.is_absolute():
+            msg = f"Namelist path {config_src} must be relative with respect to config file."
```
agoscinski, line 37:
```suggestion
            msg = f"Script 'src' {config_src} must be relative with respect to config file."
```

src/sirocco/core/_tasks/shell_task.py
```diff
@@ -1,10 +1,45 @@
...
+        src = config_src if config_rootdir is None else (config_rootdir / config_src)
+        if not src.exists():
+            msg = f"Namelist in path {src} does not exist."
+            raise FileNotFoundError(msg)
+        if not src.is_file():
+            msg = f"Namelist in path {src} is not a file."
```
agoscinski, line 43:
```suggestion
            msg = f"Script in path {src} is not a file."
```

GeigerJ2 (approved):
Minor nitpicks, then this should be good to go, I think

tests/cases/large/data/config.txt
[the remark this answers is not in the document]
agoscinski: We reintroduce this with #184 because there we introduce two separate variable names and we can use the relative path in the representation

PR #175, opened 2025-06-10
Support for `IconTask` to produce `GeneratedData` to `ShellTask`

agoscinski:
This allows to add output data to `IconTask` and to pass it a `ShellTask`. Note that this PR does enable to pass `GeneratedData` to `IconTask` as input. If this is also needed, we can do it in a separate PR.

TODO:
- [x] workgraph interprets dots in name as nested namespace, this seems to be bug already present in main, probably best to fix it in separate PR
--> Talked with Xing, this is the expected behavior of ShellJob. The retrieval of generated files is not part of workgraph but part of ShellJob. We use `metadata.options.additional_retrieve_list` to imitate this behavior in IconCalculation
- [ ] Create issue from https://github.com/C2SM/Sirocco/pull/175#discussion_r2138302746


tests/cases/small-icon/config/config.yml
```diff
@@ -31,13 +31,19 @@ cycles:
                   port: finish_status
               - restart:
                   port: latest_restart_file
+              - atmo_log
   - lastly:
+      cycling:
+        start_date: '2026-03-01T00:00' # TODO change to 05-06 later
+        stop_date: '2026-04-01T00:00'
+        period: P1M
```
agoscinski, line 39: This is a workaround because there is some error handling in one-of-cycles that prevents using dates here. The original input was
```yaml
        - cleanup:
            wait_on:
              - icon:
                  target_cycle:
                    date: 2026-05-01T00:00
            inputs:
              - atmo_log:
                  when:
                    at: 2026-05-01T00:00
                  target_cycle:
                    lag: ['P0M']
                  port: None
```
but produces
```python

    def is_active(self, date: datetime | None) -> bool:
        if date is None:
            msg = "Cannot use a when.at specification in a one-off cycle"
>           raise ValueError(msg)
E           ValueError: Cannot use a when.at specification in a one-off cycle

src/sirocco/parsing/when.py:27: ValueError
```
Could you solve this in a separate PR @leclairm or maybe you have in mind how this should look. Because I need to specify a date due to another error
```python
    def iter_from_cycle_spec(self, spec: TargetNodesBaseModel, ref_coordinates: dict) -> Iterator[GRAPH_ITEM_T]:                                                                                         
        # Check date references                                                                                                                                                                          
        if "date" not in self._dims and isinstance(spec.target_cycle, DateList | LagList):                                                                                                               
            msg = f"Array {self._name} has no date dimension, cannot be referenced by dates"                                                                                                             
            raise ValueError(msg)                                                                                                                                                                        
        if "date" in self._dims and ref_coordinates.get("date") is None and not isinstance(spec.target_cycle, DateList):
            msg = f"Array {self._name} has a date dimension, must be referenced by dates"           
>           raise ValueError(msg)                                                                   
E           ValueError: Array atmo_log has a date dimension, must be referenced by dates            
                                                  
src/sirocco/core/graph_items.py:216: ValueError
```

GeigerJ2 (approved):
Thanks, @agoscinski! Great work!

PR #176, opened 2025-06-11
Add minimal CLI interface using typer and rich

GeigerJ2:
Currently, provides the `sirocco`:
- `verify`: verify workflow spec in YAML file
- `represent`: pretty-printing text-based representation
- `visualize`: generating interactive SVG file
- `run`: run WG blockingly via AiiDA
- `submit`: submit to AiiDA daemon

Still need to fix `fmt` and `types:check` and update docs.


tests/test_cli.py
```diff
@@ -0,0 +1,226 @@
...
+        """Test the submit command with wait option."""
+        with patch("sirocco.cli._prepare_aiida_workgraph") as mock_prepare:
+            mock_aiida_wg = Mock()
+            mock_aiida_wg._core_workflow.name = "test_workflow"  # noqa: SLF001
+            mock_result = Mock()
+            mock_result.pk = 12345
+            mock_aiida_wg.submit.return_value = mock_result
+            mock_prepare.return_value = mock_aiida_wg
+
+            result = runner.invoke(app, ["submit", str(sample_workflow_file), "--wait", "--timeout", "120"])
```
agoscinski, line 169: in aiida v2.7.0 we will remove wait option (and now it is any not working)
```suggestion
            result = runner.invoke(app, ["submit", str(sample_workflow_file),  "--timeout", "120"])
```

src/sirocco/cli.py
```diff
@@ -0,0 +1,242 @@
...
+    add_completion=True,
+)
+
+# Create a Rich console instance for printing
+console = Console()
+
+
+def _prepare_aiida_workgraph(workflow_file: Path) -> AiidaWorkGraph:
+    """Helper to prepare AiidaWorkGraph from workflow file."""
+    load_profile()
```
agoscinski, line 28: this can also fail but is not in the try catch block to wrap the message

agoscinski (changes requested):
I am mostly fine the PR, just some minor revision

I am a bit worried by including emoji we cause problems in older terminals that are not supporting it. But since this package is only supported on newer HPC environments, it should be not a problem.

I do not see any reason to not use unittest but just want to make aware that this is more the pytest way to make mocks and monkeypatch `https://docs.pytest.org/en/stable/how-to/monkeypatch.html`

Improvements for future PRs:
- [ ] `sirocco` prints the help message instead of an error (same for every subcommand)
- [ ] How to make the aiida setup transparent (e.g. starting daemon, or that we require rabbitmq for submit). We can assume that everything is setup for endusers but the error messages still need to be understandable if some setup failed.

src/sirocco/cli.py
GeigerJ2: Good catch!

tests/test_cli.py
```diff
@@ -0,0 +1,226 @@
...
+
+        # Mock the entire submit process to avoid all validation
+        with patch("sirocco.cli._prepare_aiida_workgraph") as mock_prepare:
+            mock_aiida_wg = Mock()
+            mock_aiida_wg._core_workflow.name = "minimal_submit_test"  # noqa: SLF001
+
+            # Create a simple mock result that bypasses isinstance checks
+            mock_result = Mock()
+            mock_result.pk = 12345
+            mock_aiida_wg.submit.return_value = mock_result
```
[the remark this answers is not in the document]
GeigerJ2: Done!

tests/test_cli.py
```diff
@@ -0,0 +1,226 @@
+"""Tests for the sirocco CLI interface.
```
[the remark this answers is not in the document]
GeigerJ2: Moved it there now

agoscinski (approved):
I am not sure if it is smart to use `minimal_config_path` if we actually mock it anyway and don't check its content. Like using an empty file makes more explicit that this does not apply any operation on the file. But its really nitpicky and we might want to include some kind of testing that only happens in the CLI so approve.

PR #183, opened 2025-06-11
Add mpi, slurm and ssh to CI (using aiida-core tech stack)

GeigerJ2:


GeigerJ2:
While the setup of the container with SLURM and connection via SSH currently works, the next issue we'd need to solve is having a running ICON executable in the container. The missing files for the tests we could easily copy into the container, but getting ICON there doesn't seem straightforward. It might be of more use (for now), to just use the CSCS-CI infrastructure, as we have a working ICON executable there.

PR #184, opened 2025-06-12
Rename `src` to `path` and `relpath`

agoscinski:
Based on PR #172 

Rename `src` to `relpath` if it is required to be a relative path (in case of `NamelistFile` and `ShellTask`).  Rename `src` to `path` if it is required to be an absolute path (in case of `AvailableData`). By this change we can include the `relpath` into the representation. We also include `IconTask.bin` in the representation as this is absolute.


src/sirocco/parsing/yaml_data_models.py
```diff
@@ -367,90 +380,69 @@ class ConfigShellTask(ConfigBaseTask, ConfigShellTaskSpecs):
...
-    bin: Path = field(repr=False)
-
-    @field_validator("bin")
-    @classmethod
-    def validate_is_absolute(cls, value: Path) -> Path:
-        if not value.is_absolute():
-            msg = "The field 'bin' must be absolute path."
-            raise ValueError(msg)
-        return value
+    bin: Annotated[Path, AfterValidator(is_absolute_path)] = field(repr=True)
```
GeigerJ2, line 445: `AfterValidator` being used here on a `dataclass` rather than pydantic model, which shouldn't do anything. Either we make `ConfigIconTaskSpecs` a pydantic model, as well, move the validation to the `ConfigIconTask` and `IconTask` subclasses, or implement custom validation code here for the dataclass?

GeigerJ2 (commented):
LGTM. Just some minor comments on the implementation. With all the conflicts, we'll probably have to cherry-pick to main? Will approve once they are resolved. Thanks, @agoscinski!

src/sirocco/parsing/yaml_data_models.py
agoscinski: It  works since we inherit from `ConfigIconTaskSpecs` with `ConfigIconTask` which is a pydantic base model. But it is true to have the validation in the dataclass is confusing. I put the validator in the Annotated to reuse the validator for multiple classes. But I think its better to put back the validator to `ConfigIconTask`

PR #186, opened 2025-06-12
Use `SinglefileData` when data uses computer with local transport type

agoscinski:
To copy data to a remote machine we cannot use `RemoteData` but have to use `SinglefileData`.


GeigerJ2 (approved):
LGTM

PR #187, opened 2025-06-12
Remove remaining `field` and introduce `extra='forbid'` to yaml models

agoscinski:
In PR #164 (commit ec8d3750) The `type` tag was removed but it was missed to remove it from the config for the test case small-icon

This flag disallows fields that are not defined in the schema. We have to do an exception for the `ConfigShellTask` and `ConfigIconTask` because these use a field as discriminator which is detected by the flag.
    
Removes plugin class var from `ConfigIconTask` as it is present in `ConfigIconTaskSpecs`.


GeigerJ2 (approved):
Nice catch!

PR #190, opened 2025-06-17
Implement MPI and scheduler options

agoscinski:
Copy mpi options from PR #183

We test these options in the ci-integration with local transport and direct scheduler.

TODO:
- [x] put remote localhost in different PR


GeigerJ2 (approved):
Nice work, @agoscinski, very clean! Thanks a lot!

PR #202, opened 2025-06-27
Enable testing for py310 py311 and py313 in CI

agoscinski:
Our code fails for <py3.12. Somehow hatch runs on different python version. I don't know where it finds executable from. But see the CI from latest run on main for py3.11
image

I needed to add to the test matrix more versions and changed the default to use system python version. The fixes that need to be done in the code are marginal.

### DIFF
--- a/src/sirocco/core/graph_items.py
+++ b/src/sirocco/core/graph_items.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import enum
 from dataclasses import dataclass, field
 from itertools import chain, product
-from typing import TYPE_CHECKING, Any, ClassVar, Self, TypeVar, cast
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -101,18 +101,19 @@ path = "src/sirocco/__init__.py"
 
 [tool.hatch.envs.default]
 installer = "uv"
-python = "3.12"
+python = "python"
```
GeigerJ2, line 104: Fix this to something static, e.g., 3.10

agoscinski, line 104:
```suggestion
python = "3.12"
```

agoscinski: we agreed to use py3.12

PR #210, opened 2025-07-15
Add `wrapper_script` field to `IconTask`

GeigerJ2:
The `wrapper_script` option is only implemented for the `IconTask`, as only the `IconCalculation` exposes it as an input, while `ShellTask`s don't.

Currently, I don't enforce any restriction on the path for the `wrapper_script`, be it relative or absolute. This is because typically, one would expect that users have their `wrapper_script`s in the same tree as their Sirocco config YAML, so the path would be relative. However, we also provide scripts in aiida-icon (e.g., `todi_cpu.sh`), so to allow people to use those, we must also allow absolute paths.

Further, if no `wrapper_script` is being set, `todi_cpu.sh` from aiida-icon is currently being used as the default.


src/sirocco/core/_tasks/icon_task.py
```diff
@@ -116,3 +121,34 @@ def build_from_config(cls: type[Self], config: models.ConfigTask, config_rootdir
...
+        return resolved_path
+
+    def get_wrapper_script_aiida_data(self) -> aiida.orm.SinglefileData | None:
+        """Get AiiDA SinglefileData for wrapper script if configured"""
+        if self.wrapper_script is not None:
+            return aiida.orm.SinglefileData(str(self.wrapper_script))
+        return self._get_default_wrapper_script()
+
+    def _get_default_wrapper_script(self) -> aiida.orm.SinglefileData | None:
+        """Get default wrapper script based on task type"""
```
agoscinski, line 145: So far I tried to avoid to have any aiida related things in core. So I would move this to workgraph module

agoscinski (changes requested):
I think for testing we need an actual `dummpy_wrapper.sh` (maybe it was missed when pushing). We need some test in unit_test/workgraph that checks if the wrapper script contains the actual file. Also I would move the aiida related code to workgraph.py

src/sirocco/core/_tasks/icon_task.py
GeigerJ2: Fine, will do! I also asked myself when adding the `aiida` import there, if that's OK

src/sirocco/core/_tasks/icon_task.py
```diff
@@ -52,6 +52,9 @@ def __post_init__(self):
             raise ValueError(msg)
         self._model_namelist = model_namelist
 
+        if self.wrapper_script is not None:
```
agoscinski, line 55: We need to allow that it is absolute so we can reference as default the script in aiida-icon

agoscinski (changes requested):
### Future PR
- [ ] cache core representation by creation a session fixture for workgraph module

src/sirocco/core/_tasks/icon_task.py
agoscinski: okay wrong comment, since we do it in workgraph level this does not restrict it. We just do it to be flexible.

agoscinski (approved):

PR #217, opened 2025-07-25
Apply `common_alps_setup`

GeigerJ2:


pyproject.toml
```diff
@@ -34,7 +34,8 @@ dependencies = [
   "pydantic",
   "ruamel.yaml",
   "aiida-core>=2.5",
-  "aiida-icon>=0.4.0",
+  # "aiida-icon>=0.4.0",
+  "aiida-icon @ git+https://github.com/GeigerJ2/aiida-icon.git@c6590f434c899921dd4f019348d920e5ef1a5aca",
```
GeigerJ2, line 38: Current pin to avoid exception from lambda default of `mpirun_extra_params`

tests/cases/APE_R02B04/config/config.yml
```diff
@@ -49,29 +49,32 @@ tasks:
       namelists:
         - ./ICON/icon_master.namelist
         - ./ICON/NAMELIST_exclaim_ape_R02B04
+      mpi_cmd: "srun -n {tot_num_mpiprocs} --ntasks-per-node {num_mpiprocs_per_machine}"
```
GeigerJ2, line 52: Check which placeholders to use `tot_num_mpiprocs` from aiida, or ours

src/sirocco/workgraph.py
```diff
@@ -349,18 +348,23 @@ def _create_icon_task_node(self, task: core.IconTask):
             buffer.seek(0)
             builder.model_namelist = aiida.orm.SinglefileData(buffer, task.model_namelist.name)
 
+        from aiida_icon.site_support.cscs.todi import common_alps_setup
+
+        common_alps_setup(builder)
```
agoscinski, line 353: This should be moved to the new `run_env` so it can be overwritten by user then we can specify it in small-icon and make the test work

PR #244, opened 2026-01-22
port mapping directly in config format

### LINKED ISSUE #241: Invert port/data order in format
switch from
```python
cycles:
  - cycle_1:
      tasks:
        - task_1:
            inputs:
              - data_1:
                  port: port_name_in_1
              - data_2:
                  port: port_name_in_1
              - data_3:
                  port: port_name_in_2
            outputs:
              - data_4:
                  port: port_name_out_1
              - data_5:
                  port: port_name_out_1
```
to
```python
cycles:
  - cycle_1:
      tasks:
        - task_1:
            inputs:
              port_name_in_1:
                - data_1
                - data_2
              port_name_in_2:
                - data_3
            outputs:
              port_name_out_1:
                - data_4
                - data_5
```
- This would make ports non-optional also for outputs (users have to create a dummy port if necessary) which help users to integrate the port concept
- It makes the format more user-friendly, typically for ports with multiple data nodes
- It mimics the internal structure we have for inputs and outputs which are `dict`s mapping ports to data nodes (using a `None` port if necessary)

leclairm:
Invert order of port and data in config file.
As a consequence ports become compulsory for outputs as well as inputs.

### DIFF
--- a/src/sirocco/core/graph_items.py
+++ b/src/sirocco/core/graph_items.py
@@ -142,7 +142,7 @@ class Task(ConfigBaseTaskSpecs, GraphItem):
     cycle: Cycle = field(init=False, repr=False)
 
     inputs: dict[str, list[Data]] = field(default_factory=dict)
-    outputs: dict[str | None, list[GeneratedData]] = field(default_factory=dict)
+    outputs: dict[str, list[GeneratedData]] = field(default_factory=dict)
     wait_on: list[Task] = field(default_factory=list)
     waiters: list[Task] = field(default_factory=list, repr=False)
     parents: list[Task] = field(default_factory=list, repr=False)
@@ -153,7 +153,7 @@ class Task(ConfigBaseTaskSpecs, GraphItem):
     def __post_init__(self):
         GraphItem.__post_init__(self)
         if set(self.inputs.keys()).intersection(self.outputs.keys()):
-            msg = "port names must be unique, even between inputs and outputs"
+            msg = f"task {self.name}: port names must be unique, even between inputs and outputs.\ninputs ports: {self.inputs.keys()}\noutputs ports: {self.outputs.keys()}"
             raise ValueError(msg)
         self.run_dir = (self.config_rootdir / self.RUN_ROOT / self.label).resolve()
 
@@ -186,18 +186,6 @@ class Task(ConfigBaseTaskSpecs, GraphItem):
         datastore: Store,
         graph_spec: ConfigCycleTask,
     ) -> Task:
-        inputs: dict[str, list[Data]] = {}
-        for input_spec in graph_spec.inputs:
-            if input_spec.port not in inputs:
-                inputs[input_spec.port] = []
-            inputs[input_spec.port].extend(datastore.iter_from_cycle_spec(input_spec, coordinates))
-
-        outputs: dict[str | None, list[Data]] = {}
-        for output_spec in graph_spec.outputs:
-            if output_spec.port not in outputs:
-                outputs[output_spec.port] = []
-            outputs[output_spec.port].append(datastore[output_spec.name, coordinates])
-
         if (plugin_cls := Task.plugin_classes.get(type(config).plugin, None)) is None:
             msg = f"Plugin {type(config).plugin!r} is not supported."
             raise ValueError(msg)
@@ -207,8 +195,14 @@ class Task(ConfigBaseTaskSpecs, GraphItem):
             config_rootdir=config_rootdir,
             coordinates=coordinates,
             cycle_point=cycle_point,
-            inputs=inputs,
-            outputs=outputs,
+            inputs={
+                port: [*chain(*(datastore.iter_from_cycle_spec(data_spec, coordinates) for data_spec in data_list))]
+                for port, data_list in graph_spec.inputs.items()
+            },
+            outputs={
+                port: [datastore[output_spec.name, coordinates] for output_spec in output_data]
+                for port, output_data in graph_spec.outputs.items()
+            },
         )
 
         # Store for actual linking in link_wait_on_tasks() once all tasks are created
--- a/src/sirocco/core/graph_items.py
+++ b/src/sirocco/core/graph_items.py
@@ -173,7 +173,7 @@ class Task(ConfigBaseTaskSpecs, GraphItem):
     def output_data_nodes(self) -> Iterator[GeneratedData]:
         yield from chain(*self.outputs.values())
 
-    def output_data_items(self) -> Iterator[tuple[str | None, GeneratedData]]:
+    def output_data_items(self) -> Iterator[tuple[str, GeneratedData]]:
         yield from ((key, value) for key, values in self.outputs.items() for value in values)
 
     @classmethod
--- a/src/sirocco/core/scheduler.py
+++ b/src/sirocco/core/scheduler.py
@@ -145,24 +145,22 @@ class Slurm(Scheduler):
             header.append(f"#SBATCH --ntasks-per-node={ntasks_per_node}")
         if output_mode == "append":
             header.append("#SBATCH --open-mode=append")
+        return header
+
+    def submit_to_scheduler(self, task: Task, dependency_type: Literal["ALL_COMPLETED", "ANY", "NONE"] = "ALL_COMPLETED",) -> str:
+        submit_cmd: list[str] = ["sbatch", "--parsable"]
         if parent_ids := [parent.jobid for parent in task.parents if parent.rank >= 0]:
-            # NOTE: We can safely remove tasks with rank -1 from the parents list
-            #       in order to avoid depending on old tasks for which the scheduler
-            #       has no info anymore when restarting after a long time.
             match dependency_type:
                 case "ALL_COMPLETED":
-                    header.append("#SBATCH --dependency=afterok:" + ":".join(parent_ids))
+                    submit_cmd.append("--dependency=afterok:" + ":".join(parent_ids))
                 case "ANY":
-                    header.append("#SBATCH --dependency=afterany:" + "?afterany:".join(parent_ids))
+                    submit_cmd.append("--dependency=afterany:" + "?afterany:".join(parent_ids))
                 case "NONE":
                     pass
                 case _:
                     assert_never(dependency_type)
-        return header
-
-    def submit_to_scheduler(self, task: Task) -> str:
+        submit_cmd.append(task.SUBMIT_FILENAME)
         with ignore_env("UENV_MOUNT_LIST", "SIROCCO_UENV", "SIROCCO_VIEW"):
-            submit_cmd: list[str] = ["sbatch", "--parsable", task.SUBMIT_FILENAME]
             result = self.run_command(submit_cmd, cwd=task.run_dir)
         return result.stdout.decode().strip()
 
--- a/src/sirocco/core/scheduler.py
+++ b/src/sirocco/core/scheduler.py
@@ -47,7 +47,7 @@ class Scheduler:
         script_lines: list[str] = ["#!/bin/bash -l", ""]
 
         # Scheduler header
-        script_lines.extend(self.header_lines(task, output_mode=output_mode, dependency_type=dependency_type))
+        script_lines.extend(self.header_lines(task, output_mode=output_mode))
 
         # Some MPI environment variables for potential usage by the user defined runscript content
         script_lines.append("")
@@ -72,7 +72,7 @@ class Scheduler:
         # Submit runscript
         # ================
         (task.run_dir / task.SUBMIT_FILENAME).write_text("\n".join(script_lines))
-        task.jobid = self.submit_to_scheduler(task)
+        task.jobid = self.submit_to_scheduler(task, dependency_type=dependency_type)
 
     def add_links(self, task: Task) -> list[str]:
         link_list: list[str] = []
@@ -91,11 +91,14 @@ class Scheduler:
         self,
         task: Task,
         output_mode: Literal["overwrite", "append"] = "overwrite",
-        dependency_type: Literal["ALL_COMPLETED", "ANY", "NONE"] = "ALL_COMPLETED",
     ) -> list[str]:
         raise NotImplementedError
 
-    def submit_to_scheduler(self, task: Task) -> str:
+    def submit_to_scheduler(
+        self,
+        task: Task,
+        dependency_type: Literal["ALL_COMPLETED", "ANY", "NONE"] = "ALL_COMPLETED",
+    ) -> str:
         raise NotImplementedError
 
     def get_status(self, task: Task) -> TaskStatus:
@@ -126,7 +129,6 @@ class Slurm(Scheduler):
         self,
         task: Task,
         output_mode: Literal["overwrite", "append"] = "overwrite",
-        dependency_type: Literal["ALL_COMPLETED", "ANY", "NONE"] = "ALL_COMPLETED",
     ) -> list[str]:
         header: list[str] = [
             f"#SBATCH --output={task.STDOUTERR_FILENAME}",
@@ -147,7 +149,11 @@ class Slurm(Scheduler):
             header.append("#SBATCH --open-mode=append")
         return header
 
-    def submit_to_scheduler(self, task: Task, dependency_type: Literal["ALL_COMPLETED", "ANY", "NONE"] = "ALL_COMPLETED",) -> str:
+    def submit_to_scheduler(
+        self,
+        task: Task,
+        dependency_type: Literal["ALL_COMPLETED", "ANY", "NONE"] = "ALL_COMPLETED",
+    ) -> str:
         submit_cmd: list[str] = ["sbatch", "--parsable"]
         if parent_ids := [parent.jobid for parent in task.parents if parent.rank >= 0]:
             match dependency_type:
--- a/src/sirocco/core/scheduler.py
+++ b/src/sirocco/core/scheduler.py
@@ -2,10 +2,10 @@ import logging
 import os
 import shutil
 import subprocess
+from abc import ABC, abstractmethod
 from contextlib import contextmanager
 from dataclasses import dataclass
 from typing import Literal, assert_never
-from abc import ABC, abstractmethod
 
 from sirocco.core.graph_items import Task, TaskStatus
 
--- a/src/sirocco/core/scheduler.py
+++ b/src/sirocco/core/scheduler.py
@@ -5,6 +5,7 @@ import subprocess
 from contextlib import contextmanager
 from dataclasses import dataclass
 from typing import Literal, assert_never
+from abc import ABC, abstractmethod
 
 from sirocco.core.graph_items import Task, TaskStatus
 
@@ -25,12 +26,12 @@ def ignore_env(*args: str):
 
 
 @dataclass(kw_only=True)
-class Scheduler:
+class Scheduler(ABC):
     def submit(
         self,
         task: Task,
         output_mode: Literal["overwrite", "append"] = "overwrite",
-        dependency_type: Literal["ALL_COMPLETED", "ANY"] = "ALL_COMPLETED",
+        dependency_type: Literal["ALL_COMPLETED", "ANY", "NONE"] = "ALL_COMPLETED",
     ):
         """Submit a task"""
 
@@ -87,25 +88,29 @@ class Scheduler:
             )
         return link_list
 
+    @abstractmethod
     def header_lines(
         self,
         task: Task,
         output_mode: Literal["overwrite", "append"] = "overwrite",
     ) -> list[str]:
-        raise NotImplementedError
+        pass
 
+    @abstractmethod
     def submit_to_scheduler(
         self,
         task: Task,
         dependency_type: Literal["ALL_COMPLETED", "ANY", "NONE"] = "ALL_COMPLETED",
     ) -> str:
-        raise NotImplementedError
+        pass
 
+    @abstractmethod
     def get_status(self, task: Task) -> TaskStatus:
-        raise NotImplementedError
+        pass
 
+    @abstractmethod
     def cancel(self, task: Task) -> None:
-        raise NotImplementedError
+        pass
 
     @classmethod
     def create(cls, key: str) -> "Scheduler":
--- a/src/sirocco/parsing/yaml_data_models.py
+++ b/src/sirocco/parsing/yaml_data_models.py
@@ -445,12 +445,11 @@ class ConfigShellTaskSpecs:
         metadata={"description": ("Script file relative to the config directory.")},
     )
 
-    # TODO: move str | None in the signature to str once ports are compulsery everywhere (also for output)
-    def resolve_ports(self, input_labels: dict[str, list[str]]) -> str:
+    def resolve_ports(self, port_labels: dict[str, list[str]]) -> str:
         """Replace port placeholders in command string with provided input labels.
 
         Returns a string corresponding to self.command with "{PORT::port_name}"
-        placeholders replaced by the content provided in the input_labels dict.
+        placeholders replaced by the content provided in the port_labels dict.
         When multiple input nodes are linked to a single port (e.g. with
         parameterized data or if the `when` keyword specifies a list of lags or
         dates), the provided input labels are inserted with a separator
@@ -531,14 +530,19 @@ class ConfigShellTaskSpecs:
                 if (opt := match_obj.group("opt")) is None:
                     msg = f"Wrong port specification: {full_match}"
                     raise ValueError(msg)
-                if not input_labels.get(port):
+                if not port_labels.get(port):
                     return cmd.replace(full_match, "")
-                return cmd.replace(full_match, opt + sep.join(input_labels[port]))
+                return cmd.replace(full_match, opt + sep.join(port_labels[port]))
 
-            if port not in input_labels:
-                msg = f"The input_labels dictionnary doesn't have the {port} key. If the port has a when condition, enclose the whole dedicated command line part in square brackets e.g. command: ... [--arg={{PORT::my_port}}] ..."
+            if port not in port_labels:
+                msg = (
+                    f"The port_labels dictionary doesn't have the {port} key. "
+                    "If the port has a when condition, enclose the whole dedicated "
+                    "command line part in square brackets "
+                    "e.g. command: ... [--arg={{PORT::my_port}}] ..."
+                )
                 raise KeyError(msg)
-            return cmd.replace(full_match, sep.join(input_labels[port]))
+            return cmd.replace(full_match, sep.join(port_labels[port]))
 
         cmd = self.command
         for when_port_match in self.when_port_pattern.finditer(cmd):
--- a/src/sirocco/parsing/yaml_data_models.py
+++ b/src/sirocco/parsing/yaml_data_models.py
@@ -310,12 +310,12 @@ def make_named_model_list_converter[NAMED_BASE_T: _NamedBaseModel](
 def make_named_model_list_dict_converter[NAMED_BASE_T: _NamedBaseModel](
     cls: type[NAMED_BASE_T],
 ) -> typing.Callable[[dict[str, list[NAMED_BASE_T | str | dict]] | None], dict[str, list[NAMED_BASE_T]]]:
+    named_model_list_converter = make_named_model_list_converter(cls)
     def convert_named_model_list_dict(
         values: dict[str, list[NAMED_BASE_T | str | dict]] | None,
     ) -> dict[str, list[NAMED_BASE_T]]:
         if values is None:
             return {}
-        named_model_list_converter = make_named_model_list_converter(cls)
         return {port: named_model_list_converter(data_list) for port, data_list in values.items()}
 
     return convert_named_model_list_dict
--- a/src/sirocco/parsing/yaml_data_models.py
+++ b/src/sirocco/parsing/yaml_data_models.py
@@ -4,11 +4,11 @@ import itertools
 import logging
 import re
 import time
-import typing
+from collections.abc import Callable
 from dataclasses import dataclass, field
 from io import StringIO
 from pathlib import Path
-from typing import Annotated, Any, ClassVar, Literal, Self
+from typing import Annotated, Any, ClassVar, Literal, Self, TypeVar, overload
 
 from pydantic import (
     AfterValidator,
@@ -120,11 +120,11 @@ def list_not_empty[ITEM_T](value: list[ITEM_T]) -> list[ITEM_T]:
     return value
 
 
-@typing.overload
+@overload
 def is_absolute_path(value: None) -> None: ...
 
 
-@typing.overload
+@overload
 def is_absolute_path(value: Path) -> Path: ...
 
 
@@ -135,11 +135,11 @@ def is_absolute_path(value: Path | None) -> Path | None:
     return value
 
 
-@typing.overload
+@overload
 def is_relative_path(value: None) -> None: ...
 
 
-@typing.overload
+@overload
 def is_relative_path(value: Path) -> Path: ...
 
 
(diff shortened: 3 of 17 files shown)


src/sirocco/parsing/yaml_data_models.py
```diff
@@ -274,20 +274,13 @@ class TargetNodesBaseModel(_NamedBaseModel):
...
+class ConfigCycleTaskWaitOn(TargetNodesBaseModel): ...
 
 
-class ConfigCycleTaskOutput(_NamedBaseModel):
-    """
-    To create an instance of an output in a task in a cycle defined in a workflow file.
-    """
-
-    port: str | None = None
+class ConfigCycleTaskOutput(_NamedBaseModel): ...
```
GeigerJ2, line 285: Maybe we can use a type alias below to make the signature more readable?

```python
type NamedModelListConverter[T: _NamedBaseModel] = Callable[
    [list[T | str | dict] | None], list[T],
]
```

then it would become:
```python
def make_named_model_list_converter[NAMED_BASE_T: _NamedBaseModel](
    cls: type[NAMED_BASE_T],
) -> NamedModelListConverter[NAMED_BASE_T]:
    ...
```

src/sirocco/parsing/yaml_data_models.py
```diff
@@ -314,19 +307,33 @@ def convert_named_model_list(
     return convert_named_model_list
```
GeigerJ2, line 310: Here similarly:

```python
type NamedModelDictConverter[T: _NamedBaseModel] = Callable[
    [dict[str, list[T | str | dict]] | None],
    dict[str, list[T]],
]
```

then it would become:
```python
def make_named_model_list_dict_converter[NAMED_BASE_T: _NamedBaseModel](
    cls: type[NAMED_BASE_T],
) -> NamedModelDictConverter[NAMED_BASE_T]:
    ...
```

src/sirocco/parsing/yaml_data_models.py
```diff
@@ -439,7 +446,7 @@ class ConfigShellTaskSpecs:
     )
 
     # TODO: move str | None in the signature to str once ports are compulsery everywhere (also for output)
```
GeigerJ2:
```suggestion
```

src/sirocco/parsing/yaml_data_models.py
```diff
@@ -314,19 +307,33 @@ def convert_named_model_list(
...
 
+def make_named_model_list_dict_converter[NAMED_BASE_T: _NamedBaseModel](
+    cls: type[NAMED_BASE_T],
+) -> typing.Callable[[dict[str, list[NAMED_BASE_T | str | dict]] | None], dict[str, list[NAMED_BASE_T]]]:
+    def convert_named_model_list_dict(
+        values: dict[str, list[NAMED_BASE_T | str | dict]] | None,
+    ) -> dict[str, list[NAMED_BASE_T]]:
+        if values is None:
+            return {}
+        named_model_list_converter = make_named_model_list_converter(cls)
```
GeigerJ2: Here, the inner converter is recreated on every call to `convert_named_model_list_dict`. Since `cls` doesn't change between invocations, the result of `make_named_model_list_converter(cls)` is always the same — it should be captured once in the closure:

```python
# Before
def make_named_model_list_dict_converter[NAMED_BASE_T: _NamedBaseModel](
    cls: type[NAMED_BASE_T],
) -> ...:
    def convert_named_model_list_dict(values):
        named_model_list_converter = make_named_model_list_converter(cls)  # rebuilt every call
        return {port: named_model_list_converter(data_list) for port, data_list in values.items()}
    return convert_named_model_list_dict
```

```python
# After
def make_named_model_list_dict_converter[NAMED_BASE_T: _NamedBaseModel](
    cls: type[NAMED_BASE_T],
) -> ...:
    named_model_list_converter = make_named_model_list_converter(cls)  # built once

    def convert_named_model_list_dict(values):
        if values is None:
            return {}
        return {port: named_model_list_converter(data_list) for port, data_list in values.items()}
    return convert_named_model_list_dict
```

src/sirocco/parsing/yaml_data_models.py
```diff
@@ -520,7 +536,7 @@ def replace_port(cmd: str, match_obj: re.Match) -> str:
                 return cmd.replace(full_match, opt + sep.join(input_labels[port]))
 
             if port not in input_labels:
-                msg = f"The input_labels dictionnary doesn't have the {port} key. If the port has a when condition, enclose the whole dedicated command line part in square brackets e.g. command: ... [--arg={{PORT:my_port}}] ..."
+                msg = f"The input_labels dictionnary doesn't have the {port} key. If the port has a when condition, enclose the whole dedicated command line part in square brackets e.g. command: ... [--arg={{PORT::my_port}}] ..."
```
GeigerJ2: Typo: `dictionnary`, and what about doing implicit string concatenation to avoid one very long string?

```suggestion
                msg = (
                    f"The port_labels dictionary doesn't have the {port} key. "
                    "If the port has a when condition, enclose the whole dedicated "
                    "command line part in square brackets "
                    "e.g. command: ... [--arg={PORT::my_port}] ..."
                )
```

src/sirocco/parsing/yaml_data_models.py
```diff
@@ -439,7 +446,7 @@ class ConfigShellTaskSpecs:
     )
 
     # TODO: move str | None in the signature to str once ports are compulsery everywhere (also for output)
-    def resolve_ports(self, input_labels: dict[str | None, list[str]]) -> str:
+    def resolve_ports(self, input_labels: dict[str, list[str]]) -> str:
```
GeigerJ2: Bit of a scope-creep, but should we rename that to `port_labels` now, throughout the method?
```suggestion
    def resolve_ports(self, port_labels: dict[str, list[str]]) -> str:
```

src/sirocco/core/graph_items.py
```diff
@@ -153,7 +153,7 @@ class Task(ConfigBaseTaskSpecs, GraphItem):
     def __post_init__(self):
         GraphItem.__post_init__(self)
         if set(self.inputs.keys()).intersection(self.outputs.keys()):
-            msg = "port names must be unique, even between inputs and outputs"
+            msg = f"task {self.name}: port names must be unique, even between inputs and outputs.\ninputs ports: {self.inputs.keys()}\noutputs ports: {self.outputs.keys()}"
```
GeigerJ2, line 156: Implicit string concatenation to make it more readable?
```suggestion
            msg = (
                f"task {self.name}: port names must be unique, even between inputs and outputs.\n"
                f"inputs ports: {self.inputs.keys()}\n"
                f"outputs ports: {self.outputs.keys()}"
            )
```

src/sirocco/core/graph_items.py
GeigerJ2: `| None` is now stale:
```suggestion
    def output_data_items(self) -> Iterator[tuple[str, GeneratedData]]:
```

src/sirocco/core/graph_items.py
```diff
@@ -207,8 +195,14 @@ def from_config(
             config_rootdir=config_rootdir,
             coordinates=coordinates,
             cycle_point=cycle_point,
-            inputs=inputs,
-            outputs=outputs,
+            inputs={
+                port: [*chain(*(datastore.iter_from_cycle_spec(data_spec, coordinates) for data_spec in data_list))]
+                for port, data_list in graph_spec.inputs.items()
+            },
```
GeigerJ2, line 201: Maybe this is more readable? Up to you!

```suggestion
            inputs={
                port: [
                    data
                    for data_spec in data_list
                    for data in datastore.iter_from_cycle_spec(data_spec, coordinates)
                ]
                for port, data_list in graph_spec.inputs.items()
            },
```

src/sirocco/core/scheduler.py
```diff
@@ -91,11 +91,14 @@ def header_lines(
...
         output_mode: Literal["overwrite", "append"] = "overwrite",
-        dependency_type: Literal["ALL_COMPLETED", "ANY", "NONE"] = "ALL_COMPLETED",
     ) -> list[str]:
         raise NotImplementedError
 
-    def submit_to_scheduler(self, task: Task) -> str:
+    def submit_to_scheduler(
+        self,
+        task: Task,
+        dependency_type: Literal["ALL_COMPLETED", "ANY", "NONE"] = "ALL_COMPLETED",
```
GeigerJ2, line 103: This accepts also `NONE`, while `submit` on the Scheduler doesn't?
```python
        dependency_type: Literal["ALL_COMPLETED", "ANY"] = "ALL_COMPLETED",
```
Should be consistent, no?

src/sirocco/core/scheduler.py
```diff
@@ -91,11 +91,14 @@ def header_lines(
...
+    ) -> str:
         raise NotImplementedError
```
GeigerJ2: Just a note for a follow-up, rather than manually raising `NotImplementedError` here, we could make this a proper abstract base class, by inheriting from `abc.ABC`, and adding the `@abstractmethod` decorator to obtain this behavior automatically.

src/sirocco/core/scheduler.py
```diff
@@ -145,24 +147,26 @@ def header_lines(
             header.append(f"#SBATCH --ntasks-per-node={ntasks_per_node}")
         if output_mode == "append":
             header.append("#SBATCH --open-mode=append")
+        return header
+
+    def submit_to_scheduler(
```
GeigerJ2, line 157: Nit: Seems like a bit of an unrelated change. But OK for me to keep it as part of this PR.

tests/cases/aquaplanet/config/config.yml
```diff
@@ -2,7 +2,7 @@ name: AQUAPLANET
 start_date: &root_start_date "2000-01-01T00:00:00"
 stop_date: &root_stop_date "2000-01-01T00:10:00"
 front_depth: 1
-cycles: # <- GRAPH DEFINITION
+cycles:
```
GeigerJ2, line 5: Hehe, good. Cleanup from the changes I did for the all-hands presentation (and that I overlooked to clean up myself 😅)

tests/cases/large/config/config.yml
```diff
@@ -135,7 +120,6 @@ tasks:
       uenv: icon/25.2:v3
   - postproc_1:
       plugin: shell
-      computer: "{{ SIROCCO_COMPUTER }}"
       path: scripts/main_script_ocn.sh
       command: "bash main_script_ocn.sh {PORT::None}"
```
GeigerJ2: Hm, this didn't raise in the tests?

```suggestion
      command: "bash main_script_ocn.sh {PORT::dummy}"
```

EDIT: `PORT::None` is no longer a valid port name — the cycle definition uses `dummy: [stream_1]`, so this should be `{PORT::dummy}`. This would cause a `KeyError` at runtime when `resolve_ports` is called during task submission.

It didn't get caught because the CI tests only parse the config and compare the pretty-printed workflow against the `.txt` reference — they never actually call `resolve_ports` / `runscript_lines`, so the `{PORT::...}` placeholders stay as opaque strings.

Same issue applies to `store_and_clean_2` below (`PORT::clean` should be `PORT::streams`, and `--archive{PORT::archive}` is missing `=`).

It might be worth adding a test that calls `resolve_ports` for each task across all test configs, validating that every `{PORT::...}` reference in a command actually matches a declared port name. That would have caught all three issues here (PORT::None, PORT::clean, missing `=`).

tests/cases/large/config/config.yml
```diff
@@ -153,18 +136,16 @@ tasks:
       walltime: 00:05:00
   - store_and_clean_1:
       plugin: shell
-      computer: "{{ SIROCCO_COMPUTER }}"
       path: scripts/post_clean.sh
-      command: "bash post_clean.sh {PORT::postout} {PORT::streams} {PORT::icon_input}"
+      command: "bash post_clean.sh --processed={PORT::postout} --original={PORT::streams} --input={PORT::icon_input} --archive{PORT::archive}"
```
GeigerJ2:
```suggestion
      command: "bash post_clean.sh --processed={PORT::postout} --original={PORT::streams} --input={PORT::icon_input} --archive={PORT::archive}"
```

tests/cases/large/config/config.yml
```diff
@@ -153,18 +136,16 @@ tasks:
...
       nodes: 1
       ntasks_per_node: 1
       cpus_per_task: 1
       walltime: 00:01:00
   - store_and_clean_2:
       plugin: shell
-      computer: "{{ SIROCCO_COMPUTER }}"
       path: scripts/post_clean.sh
-      command: "bash post_clean.sh --archive {PORT::archive} --clean {PORT::clean}"
+      command: "bash post_clean.sh --processed={PORT::postout} --original={PORT::clean} --archive{PORT::archive}"
```
GeigerJ2:
```suggestion
      command: "bash post_clean.sh --processed={PORT::postout} --original={PORT::streams} --archive={PORT::archive}"
```

tests/cases/aquaplanet/config/config.yml
GeigerJ2: Sorry, I had updated this to minutes when working on the aiida engine PR, as it seemed to me that was quicker.
```suggestion
        period: PT1H
```

tests/cases/aquaplanet/config/config.yml
```diff
@@ -2,7 +2,7 @@ name: AQUAPLANET
 start_date: &root_start_date "2000-01-01T00:00:00"
 stop_date: &root_stop_date "2000-01-01T00:10:00"
```
GeigerJ2: With `PT1H` period, the stop_date needs to cover at least the 4 cleanup dates (00:00–03:00). Also renaming the cycle to match.

```suggestion
stop_date: &root_stop_date "2000-01-01T04:00:00"
```

tests/cases/aquaplanet/config/config.yml
```diff
@@ -2,7 +2,7 @@ name: AQUAPLANET
...
 front_depth: 1
-cycles: # <- GRAPH DEFINITION
+cycles:
   - every_minute:
```
GeigerJ2:
```suggestion
  - every_hour:
```

tests/cases/large/config/config.yml
```diff
@@ -33,39 +31,29 @@ cycles:
...
-                  port: streams
-              - icon_input:
-                  port: icon_input
+              postout: [postout_1]
+              streams: [stream_1]
+              icon_input: [icon_input]
             outputs:
-              - stored_data_1
-              - atmo_log
+              archive: [stored_data_1]
```
GeigerJ2, line 56: Did `atmo_log` got dropped intentionally?

```suggestion
              archive: [stored_data_1, atmo_log]
```

If not, should also be added again to `data/generated`.

GeigerJ2 (commented):
OK, done with a first round of review. Have a bunch of comments, but nothing seems critical / blocking. Will have another round after the first ones are addressed

src/sirocco/core/graph_items.py
leclairm: yep, thanks

src/sirocco/core/graph_items.py
leclairm: indeed!

src/sirocco/core/graph_items.py
leclairm: I prefer the first version actually.

src/sirocco/core/scheduler.py
leclairm: True. It's a placeholder for later stuff that is not used yet but should be consistent.

src/sirocco/core/scheduler.py
leclairm: Is that possible to inherit from `abc.ABC` and have some methods concretely implemented and some not?
If yes, I'll go for it, thanks. I've already used it in other places.

src/sirocco/core/scheduler.py
leclairm: Yes, could have been another PR. And yes, I'll leave it ;-)

src/sirocco/core/graph_items.py
GeigerJ2: OK, up to you then 👍🏽

src/sirocco/core/scheduler.py
GeigerJ2: Yes! You need both: inheriting from `ABC` makes `@abstractmethod` work, and `@abstractmethod` marks which specific methods must be implemented by subclasses. Methods without the decorator work as normal concrete methods, so you can freely mix abstract and concrete methods in the same class.

The benefit over the current `raise NotImplementedError` approach: with `@abstractmethod`, Python raises `TypeError` at instantiation (i.e. you can't even create an instance of an incomplete subclass), rather than only failing when you call the method at runtime.

leclairm: All right, I'll use it then. Thanks.

src/sirocco/parsing/yaml_data_models.py
leclairm: I'm actually not sure it helps for readability. Parts of the type are used below when defining the callable so that would be more hidden with type aliases.

src/sirocco/parsing/yaml_data_models.py
leclairm: Hum... `cls` does change. `make_named_model_list_dict_converter` is called once with `ConfigCycleTaskInput` and once with `ConfigCycleTaskOutput`

src/sirocco/parsing/yaml_data_models.py
leclairm: indeed, thanks!

src/sirocco/parsing/yaml_data_models.py
leclairm: True, makes sense.

leclairm: Regarding all the `input_labels` names in `engines/aiida` I'd prefer to let you open a minor PR for it. I fear that if I do it in this one, I might break consistency between `input_labels` and other `input_xxx` you have there.

src/sirocco/parsing/yaml_data_models.py
leclairm: Agreed.
I usually leave that kind of formatting to `ruff` but apparently it leaves some freedom here. Or maybe we can put that in the `ruff` settings.

tests/cases/large/config/config.yml
leclairm: Yes, it did not make sense in this context. I need to add an output port with dynamical file name for all model logs of an icon task but that requires first the `core.Task.components` level I'm working on.

src/sirocco/parsing/yaml_data_models.py
GeigerJ2: Verified against `src/sirocco/parsing/yaml_data_models.py:286-321` and call sites at lines 331/335.

You're right that `cls` differs between the two factory calls, but the suggestion targets what happens *inside* each closure, not across them. Let me walk through the full call chain.

The outer factory `make_named_model_list_dict_converter` (line 310) is called exactly twice, at class definition time of `ConfigCycleTask`:

```python
# yaml_data_models.py, lines 329-336
class ConfigCycleTask(_NamedBaseModel):
    inputs: Annotated[
        dict[str, list[ConfigCycleTaskInput]],
        BeforeValidator(make_named_model_list_dict_converter(ConfigCycleTaskInput)),   # line 331
    ] = {}
    outputs: Annotated[
        dict[str, list[ConfigCycleTaskOutput]],
        BeforeValidator(make_named_model_list_dict_converter(ConfigCycleTaskOutput)),  # line 335
    ] = {}
```

Each call returns a `convert_named_model_list_dict` closure (line 313) with `cls` baked in. Call these closure_A (`cls=ConfigCycleTaskInput`) and closure_B (`cls=ConfigCycleTaskOutput`). These are independent — the suggestion doesn't propose merging them.

Now, the inner factory `make_named_model_list_converter` (line 286) is called inside the closure, on line 318:

```python
# yaml_data_models.py, lines 310-321
def make_named_model_list_dict_converter[NAMED_BASE_T: _NamedBaseModel](
    cls: type[NAMED_BASE_T],
) -> ...:
    def convert_named_model_list_dict(
        values: dict[str, list[NAMED_BASE_T | str | dict]] | None,
    ) -> dict[str, list[NAMED_BASE_T]]:
        if values is None:
            return {}
        named_model_list_converter = make_named_model_list_converter(cls)  # line 318 — rebuilt every call
        return {port: named_model_list_converter(data_list) for port, data_list in values.items()}

    return convert_named_model_list_dict
```

Every time pydantic validates an `inputs` or `outputs` field, it calls the corresponding closure. Inside that closure, `cls` is fixed. So if pydantic parses 10 tasks, closure_A is called 10 times, each time rebuilding `make_named_model_list_converter(ConfigCycleTaskInput)` with the same `cls`. The result is identical every time — wasted work.

The fix: move line 318 above the inner `def`, so it runs once per factory call instead of once per validation:

```python
def make_named_model_list_dict_converter[NAMED_BASE_T: _NamedBaseModel](
    cls: type[NAMED_BASE_T],
) -> ...:
    named_model_list_converter = make_named_model_list_converter(cls)  # built once per factory call

    def convert_named_model_list_dict(values):
        if values is None:
            return {}
        return {port: named_model_list_converter(data_list) for port, data_list in values.items()}

    return convert_named_model_list_dict
```

`make_named_model_list_converter(cls)` now runs twice total (once for Input, once for Output, at class definition) instead of once per field validation. Same semantics, but cheaper.

src/sirocco/parsing/yaml_data_models.py
GeigerJ2: I think the alias would still be a net win, but it's not a strong opinion. The way I see it: the types currently appear in two places — the factory return annotation and the inner function's own parameter/return annotations. These say the same thing. The alias would only shorten the promise (`-> NamedModelDictConverter[T]`), while the inner function keeps the types spelled out. So nothing is really hidden — you'd still see the concrete types where the implementation lives.

But I get the argument for keeping it all inline. Up to you, happy either way. For reference, this is what it would look like:

```python
type NamedModelListConverter[T: _NamedBaseModel] = Callable[
    [list[T | str | dict] | None], list[T]
]
type NamedModelDictConverter[T: _NamedBaseModel] = Callable[
    [dict[str, list[T | str | dict]] | None], dict[str, list[T]]
]


def make_named_model_list_converter[NAMED_BASE_T: _NamedBaseModel](
    cls: type[NAMED_BASE_T],
) -> NamedModelListConverter[NAMED_BASE_T]:
    def convert_named_model_list(
        values: list[NAMED_BASE_T | str | dict] | None,
    ) -> list[NAMED_BASE_T]:
        ...

    return convert_named_model_list


def make_named_model_list_dict_converter[NAMED_BASE_T: _NamedBaseModel](
    cls: type[NAMED_BASE_T],
) -> NamedModelDictConverter[NAMED_BASE_T]:
    named_model_list_converter = make_named_model_list_converter(cls)

    def convert_named_model_list_dict(
        values: dict[str, list[NAMED_BASE_T | str | dict]] | None,
    ) -> dict[str, list[NAMED_BASE_T]]:
        ...

    return convert_named_model_list_dict
```

src/sirocco/parsing/yaml_data_models.py
GeigerJ2: Ruff's formatter doesn't touch long strings — it can't know where to break them. E501 would flag the line, but the split is always manual.

tests/cases/large/config/config.yml
leclairm: true, I fixed it. Testing might wait a bit ...

src/sirocco/parsing/yaml_data_models.py
leclairm: Ok I see now, thanks. I overlooked that the validator is only created once at task definition and not at each instantiation.

src/sirocco/parsing/yaml_data_models.py
leclairm: Ok that's fine. I was a bit reluctant because I didn't take into account that the signature of the inner function will have to match the type alias with type checkers.

src/sirocco/core/scheduler.py
```diff
@@ -25,12 +26,12 @@ def ignore_env(*args: str):
...
 
 @dataclass(kw_only=True)
-class Scheduler:
+class Scheduler(ABC):
     def submit(
         self,
         task: Task,
         output_mode: Literal["overwrite", "append"] = "overwrite",
-        dependency_type: Literal["ALL_COMPLETED", "ANY"] = "ALL_COMPLETED",
+        dependency_type: Literal["ALL_COMPLETED", "ANY", "NONE"] = "ALL_COMPLETED",
```
GeigerJ2, line 34: Nit: `"NONE"` as a string literal is awkward when Python already has `None`. `Literal["ALL_COMPLETED", "ANY"] | None = "ALL_COMPLETED"` would be more idiomatic, and the `match` case becomes `case None: pass`. No callers need updating (the only non-default call site is `workflow.py:420` with `"ANY"`). Same signature appears on lines 103 and 160.

src/sirocco/core/scheduler.py
```diff
@@ -87,22 +88,29 @@ def add_links(self, task: Task) -> list[str]:
             )
         return link_list
 
+    @abstractmethod
     def header_lines(
         self,
         task: Task,
         output_mode: Literal["overwrite", "append"] = "overwrite",
```
GeigerJ2, line 95: Nit: Just one note here because, once I learned about `Literal`s, I was using them excessively. Here, rather than the `output_mode` `Literal`, one could just have `overwrite: bool = True`. If there's only two `Literal` members, boolean flag is usually simpler. But not blocking in any way :)

src/sirocco/parsing/yaml_data_models.py
```diff
@@ -274,25 +274,19 @@ class TargetNodesBaseModel(_NamedBaseModel):
...
 
-class ConfigCycleTaskOutput(_NamedBaseModel):
-    """
-    To create an instance of an output in a task in a cycle defined in a workflow file.
-    """
+class ConfigCycleTaskOutput(_NamedBaseModel): ...
 
-    port: str | None = None
 
+type NamedModelListConverter[T: _NamedBaseModel] = Callable[[list[T | str | dict] | None], list[T]]
```
GeigerJ2, line 286: I realized `type` is only Python 3.12+. We should be using `TypeAlias` to be compatible with Python 3.10+ here, to satisfy the supported Python range we give in our project metadata.

However, we also use other Python 3.12+ syntax elsewhere, e.g.:  
```
  - yaml_data_models.py:116 — def list_not_empty[ITEM_T](...)
  - yaml_data_models.py:1124 — def validate_yaml_content[OBJECT_T](...)
  - graph_items.py:394 — def unique_item_list[GRAPH_ITEM_T](...)
```
So we could also bump `requires-python` to 3.12 in a follow-up.

src/sirocco/parsing/yaml_data_models.py
```diff
@@ -274,25 +274,19 @@ class TargetNodesBaseModel(_NamedBaseModel):
...
 
-def make_named_model_list_converter[NAMED_BASE_T: _NamedBaseModel](
-    cls: type[NAMED_BASE_T],
-) -> typing.Callable[[list[NAMED_BASE_T | str | dict] | None], list[NAMED_BASE_T]]:
+
+def make_named_model_list_converter[NAMED_BASE_T: _NamedBaseModel](cls: type[NAMED_BASE_T]) -> NamedModelListConverter:
```
GeigerJ2, line 289: These should be `-> NamedModelListConverter[NAMED_BASE_T]` and `-> NamedModelDictConverter[NAMED_BASE_T]`. Without the type parameter, the return annotation erases the generic binding — a type checker sees "returns a converter for any `_NamedBaseModel`" instead of "returns a converter for the specific `NAMED_BASE_T` you passed in via `cls`." The whole point of the type alias was to preserve this information.

src/sirocco/parsing/yaml_data_models.py
```diff
@@ -314,19 +308,37 @@ def convert_named_model_list(
     return convert_named_model_list
 
 
+type NamedModelDictConverter[T: _NamedBaseModel] = Callable[
+    [dict[str, list[T | str | dict]] | None], dict[str, list[T]]
+]
+
+
+def make_named_model_dict_converter[NAMED_BASE_T: _NamedBaseModel](cls: type[NAMED_BASE_T]) -> NamedModelDictConverter:
```
GeigerJ2, line 316: Same here: These should be `-> NamedModelListConverter[NAMED_BASE_T]` and `-> NamedModelDictConverter[NAMED_BASE_T]`. Without the type parameter, the return annotation erases the generic binding — a type checker sees "returns a converter for any `_NamedBaseModel`" instead of "returns a converter for the specific `NAMED_BASE_T` you passed in via `cls`." The whole point of the type alias was to preserve this information.

src/sirocco/parsing/yaml_data_models.py
```diff
@@ -274,25 +274,19 @@ class TargetNodesBaseModel(_NamedBaseModel):
...
-class ConfigCycleTaskWaitOn(TargetNodesBaseModel):
-    pass
+class ConfigCycleTaskWaitOn(TargetNodesBaseModel): ...
 
 
-class ConfigCycleTaskOutput(_NamedBaseModel):
-    """
-    To create an instance of an output in a task in a cycle defined in a workflow file.
-    """
+class ConfigCycleTaskOutput(_NamedBaseModel): ...
```
GeigerJ2, line 283: Nit: docstring got lost here

src/sirocco/parsing/yaml_data_models.py
```diff
@@ -314,19 +308,37 @@ def convert_named_model_list(
...
     To create an instance of a task in a cycle defined in a workflow file.
     """
 
     inputs: Annotated[
-        list[ConfigCycleTaskInput],
-        BeforeValidator(make_named_model_list_converter(ConfigCycleTaskInput)),
-    ] = []
+        dict[str, list[ConfigCycleTaskInput]],
+        BeforeValidator(make_named_model_dict_converter(ConfigCycleTaskInput)),
+    ] = {}
```
GeigerJ2, line 337: Nit: Mutable default values. Pydantic deep-copies these per instance, so functionally safe. But style-wise (and if `ruff` B006 is ever enabled): `Field(default_factory=dict)` / `Field(default_factory=list)` would be more explicit. Same applies to the pre-existing `parameters: ... = {}` on `TargetNodesBaseModel:274`. Not introduced by this PR, just adjacent.

GeigerJ2 (approved):
Thanks for addressing all my previous comments, @leclairm! I've added a few nitpicks, but nothing blocking. Good to get merged!

src/sirocco/parsing/yaml_data_models.py
leclairm: true!

src/sirocco/core/scheduler.py
leclairm: Thanks for pointing it. I already wrapped my head around this a few times and I'm not fully sure if it should be changed. I'll leave it for later (if ever :smile: )

src/sirocco/parsing/yaml_data_models.py
leclairm: I think we decided it was fine for type annotations cause it has no consequences at runtime. If tests are passing, I guess somehow this also gets ignored at runtime for python < 3.12 ?
Anyway, thanks and I'll also keep that for later.

src/sirocco/parsing/yaml_data_models.py
leclairm: I removed it on purpose as the other ones above don't have a docstring either I didn't find it very helpful. That doesn't mean we couldn't put helpful docstrings everywhere!

src/sirocco/core/scheduler.py
leclairm: I started with a boolean. Then ruff complained about boolean arguments. I guess I could make it kw arg only and not positional.

src/sirocco/parsing/yaml_data_models.py
leclairm: Yes, we discussed that already with @agoscinski and decided default empty list/dict are of for pydantic classes.

src/sirocco/core/scheduler.py
GeigerJ2: Huh, interesting, why would ruff complain about booleans? 😅

